PR #41, opened 2024-04-10
Review and fix config interface, schema and default values

colinrotherham:


designer/server/src/config.ts
```diff
@@ -1,130 +1,167 @@
...
-  port: process.env.PORT,
-  env: process.env.NODE_ENV,
-  appPathPrefix: process.env.APP_PATH_PREFIX || '/forms-designer',
-  previewUrl:
-    process.env.PREVIEW_URL || 'http://dev.cdp-int.defra.cloud/forms-runner/', // TODO set this to localhost and pull env vars from CDP
-  publishUrl:
-    process.env.PUBLISH_URL || 'http://dev.cdp-int.defra.cloud/forms-runner/', // TODO set this to localhost and pull env vars from CDP
-  persistentBackend: process.env.PERSISTENT_BACKEND || 'preview',
+const config: Config = {
+  port: parseInt(PORT),
```
davidjamesstone, line 120: Better to let `joi` do the type coercion from ENVAR strings (happens during `schema.validate`).
It might have to stay for while though to keep TS happy?

designer/server/src/config.ts
```diff
@@ -1,130 +1,167 @@
...
+  logLevel: LOG_LEVEL as Config['logLevel'],
+  phase: PHASE as Config['phase'],
+  footerText: FOOTER_TEXT,
+  isProduction: NODE_ENV === 'production',
+  isDevelopment: !['production', 'test'].includes(NODE_ENV),
+  isTest: NODE_ENV === 'test',
+  lastCommit: LAST_COMMIT ?? LAST_COMMIT_GH,
+  lastTag: LAST_TAG ?? LAST_TAG_GH,
+  sessionTtl: SESSION_TTL
+    ? parseInt(SESSION_TTL)
```
davidjamesstone, line 139: Same `parseInt` here.

Also, defaults/fallback values like this are better applied using `joi.default()` for consistency.

TBH I try and avoid config default values - they can become troublesome.

PR #57, opened 2024-04-16
Remove `@defra/forms-queue-model`

colinrotherham:
The queue service is flagged off in **forms-runner** via `enableQueueService`

Since we're likely not going to be using MySQL migrations via Prisma, I've opened this PR for discussion


alexluckett:
No objection

colinrotherham:
Let's get this merged then, it's no longer referenced in **forms-runner** since:

* https://github.com/DEFRA/forms-runner/pull/36

PR #64, opened 2024-04-17
Load the form editor content from the form manager API

alexluckett:
- Removes the default three-page view bundled with the form designer. Instead, it retrieves the form definition from the forms-manager API and displays the content on screen
- Temporarily removes form definition pushes to the server as we don't yet have a deployment mechanism in forms-manager. I have commented this code out rather than removing it, since it may be used in work this sprint or next and won't sit around for long.
- Form Builder has been removed as a nav bar link, replaced with the new Forms Library page
- An Edit button has been added to each row in the Forms Library, so that the user can load the editor for each form.


designer/server/src/plugins/routes/api.test.ts
```diff
@@ -140,91 +86,93 @@ describe('Server API', () => {
...
-
-    // When
-    const result = await server.inject<{ err: Error }>(options)
-
-    // Then
-    expect(result.statusCode).toBe(401)
-    expect(result.result?.err.message).toBe('Error in persistence service')
-  })
+  // TODO re-enable once the forms manager supports this
+  // test('Schema validation failures should return 401', async () => {
```
colinrotherham, line 90: Mind using `test.skip` then we always know how many we still need to come back and fix?

designer/server/src/plugins/routes/api.ts
```diff
@@ -20,68 +20,64 @@ export const getFormWithId: ServerRoute = {
...
-      let formJson = newFormJson
       try {
         const response = await getPublished(id)
-        const { values } = JSON.parse(response)
+        const formJson = JSON.parse(response)
 
-        if (values) {
-          formJson = values
-        }
+        return h.response(formJson).type('application/json')
```
colinrotherham, line 27: Is calling `.type('application/json')` optional for existing objects?

Or maybe `JSON.parse()` is optional for already-stringified JSON?

```suggestion
        return h.response(formJson).type('application/json')
```

alexluckett: nah not required, it's redundant. I can remove it.

designer/server/src/common/nunjucks/context/build-navigation.js
```diff
@@ -21,7 +21,7 @@ function buildEntry(request, text, url) {
 function buildNavigation(request) {
   return [
     buildEntry(request, 'Home', ''),
-    buildEntry(request, 'Form Builder', '/app')
+    buildEntry(request, 'Forms Library', '/library')
```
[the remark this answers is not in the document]
alexluckett: Prepped locally. Waiting on Dan's confirmation.

PR #108, opened 2024-04-29
Fix Docker image boot issues - missing libs

alexluckett:
Fixes the 

> Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@defra/forms-model' imported from /home/node/app/designer/server/dist/plugins/routes/api.js

error on image startup


Dockerfile
```diff
@@ -50,12 +50,16 @@ LABEL uk.gov.defra.ffc.parent-image=defradigital/node:${PARENT_VERSION}
...
-COPY --from=productionBuild /home/node/app/designer/node_modules ./designer/node_modules
-COPY --from=productionBuild /home/node/app/designer/package.json ./designer/package.json
+COPY --from=productionBuild --chown=node:node /home/node/app/packag*.json ./
+
+COPY --from=productionBuild --chown=node:node /home/node/app/model/package.json ./model/
+COPY --from=productionBuild --chown=node:node /home/node/app/model/dist ./model/dist
+
+COPY --from=productionBuild --chown=node:node /home/node/app/designer/package.json ./designer/package.json
+COPY --from=productionBuild --chown=node:node /home/node/app/designer/client/dist ./designer/client/dist
+COPY --from=productionBuild --chown=node:node /home/node/app/designer/server/dist ./designer/server/dist
```
colinrotherham, line 60: Presume we need the full path to support `--from` because **WORKDIR** doesn't cut it here?

```suggestion
COPY --from=productionBuild --chown=node:node ./packag*.json ./

COPY --from=productionBuild --chown=node:node ./model/package.json ./model/
COPY --from=productionBuild --chown=node:node ./model/dist ./model/dist

COPY --from=productionBuild --chown=node:node ./designer/package.json ./designer/package.json
COPY --from=productionBuild --chown=node:node ./designer/client/dist ./designer/client/dist
COPY --from=productionBuild --chown=node:node ./designer/server/dist ./designer/server/dist
```

colinrotherham (approved):

Dockerfile
colinrotherham: Yeah that's what used to happen across layers, hoped they'd fixed it but thanks for checking

PR #120, opened 2024-05-01
Apply valid organisations on the schema

davidjamesstone:


designer/server/src/models/create.js
```diff
@@ -22,6 +24,21 @@ export function titleViewModel(metadata, validation) {
...
+ * Text and Value Item
+ * @typedef {object} Item
+ * @property {string} text - The item text
+ * @property {string} value - The item value
+ */
+
+/**
+ * @type {Item[]}
+ */
+const organisationItems = organisations.map((organisation) => ({
```
davidjamesstone, line 37: I tried all sorts to remove the squiggles. Seemed like it should be easy. Please help!

designer/server/src/routes/create.js
```diff
@@ -126,6 +126,7 @@ export default [
         payload: Joi.object().keys({
           organisation: organisationSchema.messages({
             'any.required': 'Select a lead organisation',
+            'any.only': 'Select a lead organisation',
             'string.max': 'Lead organisation must be 100 characters or less'
```
davidjamesstone, line 130: Superfluous

model/src/form/form-metadata/index.ts
```diff
@@ -2,8 +2,24 @@ import Joi from 'joi'
...
+  'Marine Management Organisation – MMO',
+  'Natural England',
+  'Rural Payments Agency – RPA',
+  'Veterinary Medicines Directorate – VMD'
+]
+
 export const titleSchema = Joi.string().max(250).trim().required()
-export const organisationSchema = Joi.string().max(100).trim().required()
+export const organisationSchema = Joi.string()
+  .max(100)
```
davidjamesstone, line 19: superfluous

model/src/form/form-metadata/index.ts
```diff
@@ -2,8 +2,24 @@ import Joi from 'joi'
 
 import { type FormMetadata } from '~/src/form/form-metadata/types.js'
 
+export const organisations = [
```
colinrotherham, line 5: If you found using array map caused type issues, it's because TypeScript has trust issues

Should help telling it that the array is constant and the item types won't change

```suggestion
export const organisations = /** @type {const} */ ([
```

model/src/form/form-metadata/index.ts
```diff
@@ -2,8 +2,24 @@ import Joi from 'joi'
...
+  'Animal and Plant Health Agency – APHA',
+  'Centre for Environment, Fisheries and Aquaculture Science – Cefas',
+  'Defra',
+  'Environment Agency',
+  'Forestry Commission',
+  'Marine Management Organisation – MMO',
+  'Natural England',
+  'Rural Payments Agency – RPA',
+  'Veterinary Medicines Directorate – VMD'
+]
```
colinrotherham, line 15:
```suggestion
])
```

designer/server/src/models/create.js
```diff
@@ -38,45 +55,7 @@ export function organisationViewModel(metadata, validation) {
...
-        {
-          text: 'Rural Payments Agency – RPA',
-          value: 'Rural Payments Agency – RPA'
-        },
-        {
-          text: 'Veterinary Medicines Directorate – VMD',
-          value: 'Veterinary Medicines Directorate – VMD'
-        }
-      ],
+      items: organisationItems,
```
colinrotherham, line 58: I've suggested a TypeScript fix, should be able to stick the mapping straight into here now

colinrotherham (commented):
Looks ace, I've offered a little tweak so we don't (hopefully) have to be so strict on JSDoc

Mind we approve and merge the other PR before this one?

Thanks!

davidjamesstone:
> Mind we approve and merge the other PR before this one?

NP. I'll get the type changes done and we can merge it later. Thanks 👍

model/src/form/form-metadata/index.ts
colinrotherham: Looks like this wasn't necessary, didn't have squiggles for me. ESLint is happy too

designer/server/src/models/create.js
colinrotherham: I've inlined this to `items: organisations.map()` etc

davidjamesstone: I pullled it outside so as not to incur the cost each time the form is loaded, but agree the code looks better for it inside. Probably not worth the micro-optimisation :)

Thanks for merging :+1:

colinrotherham: Ah assumed it was to apply the JSDoc (separately) to try and fix those type issues!

PR #134, opened 2024-05-07
Replace home page with forms library

colinrotherham:
This PR removes the home page and adds:

1. Redirect to forms library after sign in
2. Updates forms library to handle empty results
3. Adjusted heading sizes to match prototype


designer/server/src/routes/account/login.js
```diff
@@ -1,17 +0,0 @@
...
-    auth: {
-      mode: 'optional',
-      strategy: 'azure-oidc'
-    }
-  }
-})
-
-/**
- * @typedef {import('@hapi/hapi').ServerRoute} ServerRoute
- */
```
alexluckett, line 17: the `/login` route is required to redirect the user to the login page (`options.auth.strategy: 'azure-oidc'` ensures that).

We'd need to keep this, otherwise you'll get a HTTP 401 when you access the library page since you won't yet have a token. I'd suggest making it consistent with your other changes and update the route to call it `/sign-in`.

colinrotherham: @alexluckett Sorry, I missed a commit from the other PR https://github.com/DEFRA/forms-designer/pull/134/commits/edaebe17b97cfefd3a185e5721ba0b8df7faa92c

Do you mind if we do this instead?

It prevents the HTTP 401 when the session expires too, and authenticates again

alexluckett: Makes perfect sense with the new commit, thanks for clarifying. Using the built-in methods for the Hapi cookie plugin is definitely better.

alexluckett (approved):

PR #135, opened 2024-05-07
Update session management to use end-user ID + types

colinrotherham:
This PR adds server methods for `session.get()`, `session.set()` and `session.drop()`

I've also typed `request.auth.user` as Hapi expects it to be with all utility methods now typed

Sessions are keyed using `claims.sub` so an expired token can seamlessly re-authenticate without session loss

```js
export function createUser(credentials) {
  const claims = getUserClaims(credentials)

  if (!claims) {
    return
  }

  return /** @satisfies {UserCredentials} */ ({
    id: claims.sub,
    email: claims.email ?? '',
    displayName: claims.name ?? '',
    issuedAt: DateTime.fromSeconds(claims.iat).toUTC().toISO(),
    expiresAt: DateTime.fromSeconds(claims.exp).toUTC().toISO()
  })
}
```

**Note:** Session expiry checks are currently turned off

### DIFF
--- a/designer/server/src/common/helpers/auth/azure-oidc.js
+++ b/designer/server/src/common/helpers/auth/azure-oidc.js
@@ -2,6 +2,7 @@ import Basic from '@hapi/basic'
 import Bell from '@hapi/bell'
 import Boom from '@hapi/boom'
 import { token } from '@hapi/jwt'
+import { DateTime } from 'luxon'
 
 import config from '~/src/config.js'
 
@@ -48,22 +49,10 @@ export const azureOidc = {
               'user.read'
             ],
             profile(credentials) {
-              const { decoded } = /** @type {UserToken} */ (
-                token.decode(credentials.token)
-              )
+              const artifacts = token.decode(credentials.token)
 
-              if (!decoded.payload) {
-                throw Boom.unauthorized()
-              }
-
-              const { payload } = decoded
-
-              credentials.profile = {
-                id: payload.oid,
-                displayName: payload.name,
-                email: payload.upn ?? payload.preferred_username,
-                loginHint: payload.login_hint
-              }
+              token.verifyTime(artifacts)
+              token.verifyPayload(artifacts)
 
               return Promise.resolve()
             }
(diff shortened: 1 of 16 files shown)


designer/server/src/common/helpers/auth/azure-oidc.js
```diff
@@ -1,121 +1,140 @@
...
-        clientId: config.azureClientId,
-        clientSecret: config.azureClientSecret,
-        providerParams: {
-          redirect_uri: authCallbackUrl
-        },
-        cookie: 'bell-azure-oidc',
-        isSecure: false,
-        config: {
-          tenant: config.azureTenantId
-        }
```
colinrotherham, line 65: This wasn't used (it seems) as it's not in the config

designer/server/src/common/helpers/auth/azure-oidc.js
```diff
@@ -1,121 +1,140 @@
...
     password: 'testing', // 'secret'
-    name: 'Joe Bloggs',
-    id: '2133d32a'
+    user: {
+      id: 'dummy-id',
+      email: 'dummy@defra.gov.uk',
+      displayName: 'John Smith',
+      issuedAt: DateTime.now().minus({ minutes: 30 }).toUTC().toISO(),
+      expiresAt: DateTime.now().plus({ minutes: 30 }).toUTC().toISO()
+    }
```
colinrotherham, line 89: All users (test fixtures, basic auth logins, OpenID Connect) now typed with Hapi `UserCredentials`

designer/server/src/common/helpers/auth/get-user-session.js
```diff
@@ -1,24 +1,69 @@
...
+  let sessionId = session?.sessionId
+
+  // Fall back to OpenID Connect (OIDC) claim
+  if (!sessionId) {
+    const claims = getUserClaims(auth.credentials)
+
+    if (claims?.sub) {
+      sessionId = claims.sub
+    }
+  }
```
colinrotherham, line 25: When a user authenticates we don't know if they have a Redis session already

E.g. Maybe their token (or cookie) has expired but the Redis session hasn't

We try and grab their `sessionId` from a few places

alexluckett (approved):

alexluckett:
Side note - refresh token support will return. Naz is adding a ticket into the backlog.

PR #137, opened 2024-05-07
Remove unused `logout_hint` to fix Keycloak sign out

colinrotherham:
We were adding `authedUser.loginHint` to `?logout_hint=` (typo?) but this isn't supported anyway

* https://github.com/keycloak/keycloak/issues/21399

We should look to add `?id_token_hint=` to enable sign out redirects in future

I've removed it for now to fix sign out locally

### DIFF
--- a/designer/server/src/routes/account/sign-out.js
+++ b/designer/server/src/routes/account/sign-out.js
@@ -12,10 +12,9 @@ export default /** @satisfies {ServerRoute} */ ({
 
     const logoutBaseUrl = request.server.app.oidc.end_session_endpoint
     const referrer = request.info.referrer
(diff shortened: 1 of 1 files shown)


alexluckett (approved):
Approving since this isn't functional in the current codebase anyway so no point keeping it yet (`logout_hint` resolves to `undefined`). For the future though, this is supported by Azure Entra so we can use it.

I've created our our dev app registration, I'm hoping we can all use that and ditch Keycloak for local development. I'll keep you posted.

colinrotherham:
Sounds good. Yep, no point leaving unused things broken and we can improve another day

PR #142, opened 2024-05-08
Feature/369420 run forms from db

davidjamesstone:


model/src/form/form-metadata/index.ts
```diff
@@ -28,11 +33,31 @@ export const teamEmailSchema = Joi.string()
  * Joi schema for `FormMetadata` interface
  * @see {@link FormMetadata}
```
colinrotherham, line 34: Thanks @davidjamesstone!

Mind doing a quick review of these `@link` tags?

E.g. This one is my fault, it should have been `FormMetadataInput`

```suggestion
 * @see {@link FormMetadataInput}
```

model/src/form/form-metadata/index.ts
```diff
@@ -28,11 +33,31 @@ export const teamEmailSchema = Joi.string()
...
  */
-export const formMetadataSchema = Joi.object<FormMetadata>()
+export const formMetadataCreateSchema = Joi.object<FormMetadata>()
```
colinrotherham, line 36: But then this type generic needs updating too

How about we name the schema to match the type too?

~`formMetadataCreateSchema`~ → `formMetadataInputSchema`

```suggestion
export const formMetadataInputSchema = Joi.object<FormMetadataInput>()
```

model/src/form/form-metadata/index.ts
```diff
@@ -28,11 +33,31 @@ export const teamEmailSchema = Joi.string()
...
   .keys({
     title: titleSchema,
     organisation: organisationSchema,
     teamName: teamNameSchema,
     teamEmail: teamEmailSchema
   })
   .required()
+
+/**
+ * Joi schema for `FormMetadataState` i nterface
```
colinrotherham, line 46:
```suggestion
 * Joi schema for `FormMetadataState` interface
```

model/src/form/form-metadata/types.ts
```diff
@@ -1,5 +1,20 @@
 import { formMetadataSchema } from '~/src/form/form-metadata/index.js'
 
+/**
+ * Interface for metadata state (draft & live)
+ */
+export interface FormMetadataState {
+  /**
+   * The date the form state was created
+   */
+  createdAt: Date
```
colinrotherham, line 10: Won't these need to be strings at some point? e.g. ISO-8601 dates

So we can correctly "type" what comes in/out of storage

```
createdAt: '2024-04-22T20:47:05.335-14:00'
```

model/src/form/form-metadata/index.ts
```diff
@@ -14,7 +17,9 @@ export const organisations = [
   'Veterinary Medicines Directorate – VMD'
 ]
 
+export const idSchema = Joi.string().hex().length(24).required()
```
colinrotherham, line 20: Nice, love how this is imported into `formByIdSchema` in **forms-manager**

model/src/form/form-metadata/index.ts
davidjamesstone: Fixed with 3bbb93aa4f770999ea2292070f7183da5e99b323

model/src/form/form-metadata/index.ts
davidjamesstone: Fixed with 3bbb93aa4f770999ea2292070f7183da5e99b323

model/src/form/form-definition/types.ts
```diff
@@ -18,7 +18,7 @@ export interface Page {
   path: string
   controller?: string
   components?: ComponentDef[]
-  section: string // the section ID
+  section?: string // the section ID
```
colinrotherham, line 21: Have you used `npm link` to test the local `@defra/forms-model`?

```console
cd forms-designer
npm link --workspace model

cd ../forms-runner
npm link ../forms-designer/model

cd ../forms-manager
npm link ../forms-designer/model
```

I've merged all the ESLint type changes from https://github.com/DEFRA/forms-runner/pull/77 in **forms-runner**

So we can see if ESLint spots usage of `.section` as a now-optional string

model/src/form/form-metadata/index.ts
```diff
@@ -30,10 +31,10 @@ export const teamEmailSchema = Joi.string()
   .required()
 
 /**
- * Joi schema for `FormMetadata` interface
- * @see {@link FormMetadata}
+ * Joi schema for `FormMetadataInput` interface
+ * @see {@link FormMetadataInput}
  */
-export const formMetadataCreateSchema = Joi.object<FormMetadata>()
+export const formMetadataInputSchema = Joi.object<FormMetadata>()
```
colinrotherham, line 37: Sorry, we need to align the type too

```suggestion
export const formMetadataInputSchema = Joi.object<FormMetadataInput>()
```

model/src/form/form-metadata/index.ts
```diff
@@ -55,7 +56,7 @@ export const formMetadataStateSchema = Joi.object<FormMetadataState>().keys({
  * Joi schema for `FormMetadata` interface
  * @see {@link FormMetadata}
  */
-export const formMetadataSchema = formMetadataCreateSchema.append({
+export const formMetadataSchema = formMetadataInputSchema.append({
```
colinrotherham, line 59: Think we're supposed to tell Joi that we've appended `FormMetadata` types?

Although it might only want the appended key/values, not the whole interface

```suggestion
export const formMetadataSchema = formMetadataInputSchema.append<FormMetadata>({
```

model/src/form/form-definition/types.ts
davidjamesstone: Yes I have been using `npm link` in the three repos. Previously I had been hacking the `package.json` and pointing the dependency to a local/relative path :face_exhaling:

model/src/form/form-metadata/index.ts
davidjamesstone: That's neat.

model/src/form/form-metadata/index.ts
davidjamesstone: Fixed in 97b16a03e5760c938d8c46dfea0320a5d8d44820

model/src/form/form-metadata/index.ts
davidjamesstone: Fixed in 97b16a03e5760c938d8c46dfea0320a5d8d44820

model/src/form/form-metadata/types.ts
davidjamesstone: They come out of storage as `Date`s (mongo uses bson which has support for dates, not json).

The only time they are stringified is when they're over the wire from the forms-manager API endpoint. Running them through the Joi schema like I do [here](https://github.com/DEFRA/forms-runner/blob/1ff8c3a815125309eae405e405d63feddbc752eb/src/server/plugins/engine/services/formsService.js#L16) gets them back to `Date`s.

model/src/form/form-metadata/index.ts
davidjamesstone: Resolved with 3bbb93aa4f770999ea2292070f7183da5e99b323

PR #145, opened 2024-05-08
Add support for scopes based on AD group IDs + reject user if required scopes aren't added

alexluckett:
### Scope support
- Reads AD groups and maps them to scopes in `./designer/server/src/common/helpers/auth/azure-oidc.js`. This is then propogated to the cookie-based session and validated on `/auth/callback`.
- As a temporary measure until more fine-grained scope support is added around the app, if no scopes are assigned to the user they will be rejected from the login process and their session will be dropped. This means we can easily support extra scopes in the future (e.g. superadmins, guest readers, etc) while having the short term goal met of rejecting users with insufficient privileges.

### UI changes
- A new sign-in screen is the new landing page, replacing the previous "you are signed out" screen
- If a user has insufficient scopes, an error message is displayed on the sign-in screen. The user will never see that they are logged in, as we log them in, check their scopes, and immediately drop their session if they don't have permission.

image


designer/server/src/routes/account/auth.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const result = await server.inject(options)
+
+    const cookies = setCookieParser.parse(result.headers['set-cookie'] ?? [], {
+      map: true
+    })
+
+    expect(
+      /** @type {CookieAuthResponse} */ (cookies).userSession?.value
+    ).toBeFalsy()
```
colinrotherham, line 39: Could we ditch the extra package/types and use Jest matchers?

```suggestion
    expect(result.headers['set-cookie']).toMatchObject(
      expect.arrayContaining([expect.stringContaining('userSession=')])
    )
```

designer/server/src/views/account/sign-in.njk
```diff
@@ -0,0 +1,29 @@
...
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+
+{% block content %}
+  {% if userFailedAuthorisation %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: [
+        {
+          html: "We couldn't sign you in. Please contact the system administrator for help, <a href=\"mailto:steven.thomas@defra.gov.uk\">steven.thomas@defra.gov.uk</a>."
```
colinrotherham, line 11: Couple of tiny text + link class tweaks

```suggestion
          html: "We couldn’t sign you in. Please contact the system administrator for help, <a href=\"mailto:steven.thomas@defra.gov.uk\" class="govuk-link">steven.thomas@defra.gov.uk</a>."
```

<br>

But ideally, can we use the [`appErrorSummary()` component](https://github.com/DEFRA/forms-designer/commit/24c87625eaba1f64fa68657e91286cd88ac02ff0) wrapper I made?

```njk
{% from "error-summary/macro.njk" import appErrorSummary %}
```

I've extended it for form creation errors that come from the API, but something like this:

* https://github.com/DEFRA/forms-designer/pull/146

Otherwise we mainly need to notify screen readers of the error

## Page title
First we'll need to suffix the page title with `Error: `

```njk
{% if userFailedAuthorisation | length %}
  {% set pageTitle = "Error: " + pageTitle %}
{% endif %}
```

## HTML block `safe` filter
But good to use HTML blocks via `safe` too, here's an example via `appErrorSummary()`

```njk
  {% set errorHtml %}
    We couldn’t sign you in. Please contact the system administrator for help,
    <a href="mailto:steven.thomas@defra.gov.uk" class="govuk-link">steven.thomas@defra.gov.uk</a>
  {% endset %}

  {% if userFailedAuthorisation %}
    {{ appErrorSummary({
      errorList: [{
        html: errorHtml | safe
      }]
    }) }}
  {% endif %}
```

designer/server/src/models/account/auth.js
```diff
@@ -1,5 +1,17 @@
...
     pageTitle: 'You have signed out'
   }
 }
+
+/**
+ *  @param {boolean} userFailedAuthorisation - indicates the user has failed authorisation, but authentication was successful
+ */
+export function signInViewModel(userFailedAuthorisation) {
+  return {
+    pageTitle: `Sign in to ${config.serviceName}`,
```
colinrotherham, line 14: Thanks for moving more bits into view models 🙌

Not a blocker, but might need to customise this `pageTitle` when auth scopes prevent sign in

designer/server/src/models/account/auth.js
```diff
@@ -1,5 +1,17 @@
...
+    userFailedAuthorisation
```
colinrotherham, line 15: Not a blocker, but we could ahead and build the `errorList` array here

Making the Nunjucks view nice and generic for shared usage

designer/server/src/models/account/auth.js
```diff
@@ -1,5 +1,17 @@
+import config from '~/src/config.js'
+
 export function signedOutViewModel() {
   return {
     pageTitle: 'You have signed out'
   }
 }
+
+/**
+ *  @param {boolean} userFailedAuthorisation - indicates the user has failed authorisation, but authentication was successful
```
colinrotherham, line 10:
```suggestion
 * @param {boolean} userFailedAuthorisation - indicates the user has failed authorisation, but authentication was successful
```

designer/server/src/common/helpers/auth/azure-oidc.js
```diff
@@ -138,3 +168,8 @@ export const azureOidcNoop = {
  * @typedef {import('oidc-client-ts').OidcMetadata} OidcMetadata - OpenID Connect (OIDC) metadata
  * @typedef {import('oidc-client-ts').UserProfile} UserProfile - User profile
  */
+
+/**
+ * @typedef {object} idTokenPayload
```
colinrotherham, line 173: I'm torn by this, does the `oidc-client-ts` package have ID token types already?

I'm guessing we've hit custom Microsoft specific territory now 😣

(Does this PR break Keycloak sign in?)

designer/server/src/common/constants/rbac.js
```diff
@@ -0,0 +1,12 @@
+import config from '~/src/config.js'
+
+export const SCOPE_READ = 'read'
+export const SCOPE_WRITE = 'write'
+
+/**
+ * Maps a group a user is a member of to the scopes they have access to.
+ * @type {{[key: string]: Array<string>}}
```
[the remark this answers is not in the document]
alexluckett: Role-based access control. 

I need to rename this back to scopes - I originally had a concept of a 'role' (hence the name) but I removed it, just need to rename the file.

designer/server/src/routes/account/auth.test.js
alexluckett: Aha!! I was looking for something similar. Thank you.

designer/server/src/views/account/sign-in.njk
colinrotherham: I've merged https://github.com/DEFRA/forms-designer/pull/146 now if you'd like to rebase

designer/server/src/common/helpers/auth/azure-oidc.js
```diff
@@ -48,12 +68,22 @@ export const azureOidc = {
...
               token.verifyPayload(artifacts)
 
+              const idTokenPayload = /** @type {{payload: idTokenPayload}} */ (
+                token.decode(idToken).decoded
+              ).payload
+              const assignedScopes = getScopesForUserProfile(
+                idTokenPayload.groups ?? []
+              )
+
+              credentials.scope = assignedScopes
```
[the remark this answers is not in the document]
alexluckett: I don't think so. That function is per request, rather than when the token is received. We only need to grab their groups when the user is signing in or is refreshed. My understanding of the scope handler is where you want to elevate the user's access based on a specific request, which wouldn't be the case here.

`credentials.scope` is something I'm still trying to work out. It's definitely the right way to do it, I'm a bit puzzled though at the type errors. The [docs](https://hapi.dev/api/?v=21.3.3#-routeoptionsauthaccessscope) indicate it's a property of credentials, as does a few guides I found online.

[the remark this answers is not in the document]
colinrotherham: Probably Bell using its own bespoke `request` types?

You'll see in https://github.com/DEFRA/forms-designer/pull/145#discussion_r1595102058 that `credentials.scope` is typed correctly there?

Be good to move the code into `createUserSession()` so the compiler can check our homework

```patch
      // Create user session
      const credentials = await createUserSession(request)

      if (!credentials?.user) {
        return Boom.unauthorized()
      }
      
+     if (!credentials?.scope) {
+       // Do the flashy auth failed thing
+     }
```

designer/server/src/views/account/sign-in.njk
alexluckett: Done

designer/server/src/models/account/auth.js
alexluckett: Done

designer/server/src/models/account/auth.js
alexluckett: Done

designer/server/src/models/account/auth.js
alexluckett: Done

designer/server/src/models/account/auth.js
alexluckett: Done

alexluckett: Not done, never mind.

designer/server/src/common/constants/scopes.js
```diff
@@ -0,0 +1,12 @@
+import config from '~/src/config.js'
+
+export const SCOPE_READ = 'read'
+export const SCOPE_WRITE = 'write'
+
+/**
+ * Maps a group a user is a member of to the scopes they have access to.
+ * @type {{[key: string]: Array<string>}}
```
colinrotherham, line 8: Just linking this new file to the old thread in https://github.com/DEFRA/forms-designer/pull/145#discussion_r1595071183

designer/server/src/views/account/signed-out.njk
```diff
@@ -1,9 +1,10 @@
 {% extends 'layouts/page.njk' %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
```
colinrotherham, line 3:
```suggestion
```

designer/server/src/common/templates/layouts/page.njk
```diff
@@ -12,7 +12,13 @@
...
 {% endblock %}
 
-{% block pageTitle %}{{ pageTitle }} - {{ config.serviceName }}{% endblock %}
+{% if errorList | length %}
+  {% set errorPrefix = "Error: " %}
+{% else %}
+  {% set errorPrefix = "" %}
+{% endif %}
+
+{% block pageTitle %}{{ errorPrefix }}{{ pageTitle }} - {{ config.serviceName }}{% endblock %}
```
colinrotherham, line 21: Good idea!

It's why I had to use `pageHeading or pageTitle` to return a non-appending header!

Only problem now is the create form journey 😣

I've copied CDP and used `formErrors` _not_ `errorList` which means `pageTitle` still needs appending on those pages. Plus they use `buildErrorDetails()` which returns a slightly incompatible array

Probably best keeping the old way for now, until we can consolidate?

```njk
{% if formErrors | length %}
  {% set pageTitle = "Error: " + pageTitle %}
{% endif %}
```

colinrotherham: Top Nunjucks trick though if you want to slim this down

**Trimmed**
```
{%- -%}
{{- -}}
```

**Untrimmed**
```
{% %}
{{ }}
```

```suggestion
{% block pageTitle -%}
  {{ "Error: " if errorList | length }}{{ pageTitle }} - {{ config.serviceName }}
{%- endblock %}
```

alexluckett: Weirdly enough I tried exactly that and it broke with some super weird error, hence my slightly verbose result.

alexluckett: I'll move this bit out to a separate PR branch/PR.

designer/server/src/views/account/signed-out.njk
alexluckett: Turns out my "better" Nunjucks extension isn't better, it doesn't flag this! Thanks.

designer/server/src/models/account/auth.js
alexluckett: Done

designer/server/src/views/account/signed-out.njk
alexluckett: Done

designer/server/src/common/helpers/auth/azure-oidc.js
alexluckett: Keycloak will probably still work but isn't required anymore, I expect we'll all switch to AAD with a DefraDev account for local development. That way we'll have a more consistent environment between local-dev/dev/test/prod.

I've removed this as it's not required anymore.

designer/package.json
```diff
@@ -65,13 +65,15 @@
...
-    "resolve": "^1.22.8"
+    "resolve": "^1.22.8",
+    "set-cookie-parser": "^2.6.0"
   },
   "devDependencies": {
     "@testing-library/dom": "^9.3.4",
     "@testing-library/jest-dom": "^6.4.2",
     "@testing-library/react": "^12.1.5",
     "@testing-library/user-event": "^14.5.2",
+    "@types/set-cookie-parser": "^2.4.7",
```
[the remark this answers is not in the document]
alexluckett: Removed, not required.

designer/server/src/common/helpers/auth/azure-oidc.js
colinrotherham: Ace, thank you

I'd like to test your PR, how do I set up Keycloak to work with the new scopes?

designer/server/src/common/constants/scopes.js
colinrotherham: Sorry, that's what I was getting it

If you do `Array<string>` or `string[]` the compiler never knows what scopes are valid

If you exactly type the list of scopes, then it'll be able to error when rename or remove one

colinrotherham: In fact, can remove the `@type` entirely

alexluckett: That's a fair point - it's redundant! Removed.

designer/server/src/common/helpers/auth/azure-oidc.js
alexluckett: I'll give you a ring after standup

designer/server/src/routes/account/auth.js
```diff
@@ -19,6 +20,16 @@ export default [
         return Boom.unauthorized()
       }
 
+      if ((credentials.scope ?? []).length === 0) {
```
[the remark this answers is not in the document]
alexluckett: I'm not using `scope` from the AT at the moment, using the ID token as that's how AD has been configured so I will need to parse it separately to `artifacts.scope`. I'll pick this up as a follow-up when I do a bit of a review of the AD setup.

Cheers for flagging.

designer/server/src/models/account/auth.js
```diff
@@ -1,5 +1,27 @@
...
+
+/**
+ * @param {boolean} userFailedAuthorisation - indicates the user has failed authorisation, but authentication was successful
+ */
+export function signInViewModel(userFailedAuthorisation) {
+  const errorList = []
+
+  if (userFailedAuthorisation) {
+    errorList.push({
+      html: `We couldn’t sign you in. Please contact the system administrator for help,
```
colinrotherham, line 17: Mind double checking if this is correct?

Usually avoid [negative contractions](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#contractions-abbreviations-and-acronyms) like "couldn't"

```suggestion
      html: `We could not sign you in. Please contact the system administrator for help,
```

alexluckett: I lifted it from Figma, I'll check with Dan.

alexluckett: Changed approved by Dan. Fixed.

designer/server/src/routes/account/auth.test.js
```diff
@@ -34,6 +34,23 @@ describe('Auth test', () => {
     expect(result.headers.location).toBe('/')
   })
 
+  test('Auth checks should reject user if groups are missing present and redirect to home', async () => {
```
nayij, line 37: 'missing present' sounds a bit confusing

designer/server/jest.setup.cjs
```diff
@@ -1 +1,2 @@
 process.env.SESSION_COOKIE_PASSWORD = 'test-env-session-cookie-password'
+process.env.FORMS_EDITOR_AD_GROUP_ID = 'valid-test-group'
```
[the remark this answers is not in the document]
alexluckett: Well, I guess it's advertising for Microsoft when you think about it 😆 but agreed, no need to introduce tech-specific terminology. Done.

designer/server/src/common/constants/scopes.js
```diff
@@ -0,0 +1,11 @@
+import config from '~/src/config.js'
+
+export const SCOPE_READ = 'read'
+export const SCOPE_WRITE = 'write'
+
+/**
+ * Maps a group a user is a member of to the scopes they have access to.
+ */
+export const groupsToScopes = {
+  [config.formsEditorAdGroupId]: [SCOPE_READ, SCOPE_WRITE]
```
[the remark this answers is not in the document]
alexluckett: Done

designer/server/src/routes/account/auth.test.js
alexluckett: Done

designer/server/test/fixtures/auth.js
```diff
@@ -7,18 +10,59 @@ export const auth = {
   strategy: 'azure-oidc',
   credentials: {
```
colinrotherham, line 11: @alexluckett I'm seeing type errors on this line, can you check?

>Type '{ provider: "azure-oidc"; refreshToken: string; token: string; }' is missing the following properties from type 'AuthCredentials<UserCredentials, AppCredentials>': query, idToken, expiresIn

i.e. We've removed **query**, **idToken**, **expiresIn** and not put them back

Happy to do a quick run through and fix a few bits?

Love how you've added `token.generate()`

colinrotherham: Previously our auth fixture was "this person has already signed in" with everything in place

So it might take some wrangling with types

PR #147, opened 2024-05-09
Add unique slug validation to “Form name” step

colinrotherham:
This PR adds validation to the "Form name" step to check if it's unique

I've also reviewed and updated validation messages with @Daniel-Da-Silveira

Form validation showing form name already exists

### DIFF
--- a/designer/server/src/lib/fetch.js
+++ b/designer/server/src/lib/fetch.js
@@ -9,6 +9,10 @@ import Wreck from '@hapi/wreck'
 export async function request(method, url, options) {
   const response = await Wreck.request(method, url.href, options)
 
+  if (response.statusCode === 404) {
+    return { response, body: undefined }
+  }
+
   if (response.statusCode !== 200) {
     throw new Error(response.statusMessage)
   }
--- a/designer/server/src/routes/forms/create.js
+++ b/designer/server/src/routes/forms/create.js
@@ -1,6 +1,7 @@
 import {
   formMetadataInputSchema,
   organisationSchema,
+  slugSchema,
   teamEmailSchema,
   teamNameSchema,
   titleSchema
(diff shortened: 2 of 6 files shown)


designer/server/src/lib/fetch.js
```diff
@@ -9,6 +9,10 @@ import Wreck from '@hapi/wreck'
 export async function request(method, url, options) {
   const response = await Wreck.request(method, url.href, options)
 
+  if (response.statusCode === 404) {
+    return { response, body: undefined }
+  }
```
colinrotherham, line 14: This prevents us throwing errors and having to handle them internally

When requests to **forms-manager** throw an Error 404 we just return an empty body

(Similar to [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) where a 404 doesn't throw)

alexluckett (approved):

model/src/form/form-metadata/index.ts
```diff
@@ -21,10 +21,18 @@ export const organisations = [
 
 export const idSchema = Joi.string().hex().length(24).required()
 export const titleSchema = Joi.string().max(250).trim().required()
-export const slugSchema = Joi.string().required()
+export const slugSchema = Joi.string()
```
davidjamesstone, line 24: I don't think this processing belongs here on `slugSchema`.
If `slug` was a user inputted field, then this code is correct, but it isn't.
Slug is just a string, built from `title` using _some algorithm_ that lives in `forms-manager`. We should export that and use it here in `forms-designer` (or better still use [slugify](https://www.npmjs.com/package/slugify) or similar so we don't stumble over the probably many edge cases there are.)

Having the logic in two places feels off and open to us having a mismatch.

Also, `slugSchema` is re-used in other validation. E.g. in the `forms-runner`, if the user goes to `/my--slug`, we should take whatever `slug` param they send ("my--slug") and use it verbatim and not have joi turn it into "my-slug"

designer/server/src/routes/forms/create.js
```diff
@@ -72,10 +76,35 @@ export default [
     options: {
       validate: {
         payload: Joi.object().keys({
-          title: titleSchema.messages({
-            'string.empty': 'Enter your form name',
-            'string.max': 'Form name must be 250 characters or less'
-          })
+          title: titleSchema
+            .external(
```
davidjamesstone, line 80: I would recommend avoiding `joi.external` and do any "2nd phase" validation logic like this in the main handler.

davidjamesstone:
I was reviewing this when it got merged but I have added a couple of retrospective thoughts.

model/src/form/form-metadata/index.ts
colinrotherham: That's fair, I've set up a new `slugify()` helper in:

* https://github.com/DEFRA/forms-designer/pull/149

I was going to import `slugSchema` into the other repos but this works too

Let me know what you think

designer/server/src/routes/forms/create.js
colinrotherham: Ah can you explain why?

Only because we've got to validate the entire metadata before submission soon:

* https://github.com/DEFRA/forms-designer/pull/150

Putting it into a schema gives us a tidy way to limit error handling to `failAction()`

PR #148, opened 2024-05-09
Add author to Form Metadata create endpoint

davidjamesstone:


designer/server/src/lib/forms.js
```diff
@@ -29,13 +29,13 @@ export async function get(slug) {
 
 /**
  * Create form
- * @param {FormMetadataInput} metadata
+ * @param {FormMetadataAuthorInput} input
```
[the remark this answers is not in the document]
colinrotherham: Alternatively we can rename the param to `user` and keep the Hapi types?

```suggestion
 * @param {FormMetadataInput} metadata
 * @param {UserCredentials} user
```

<br>

They'd come from here:

```js
/**
 * @typedef {import('@hapi/hapi').UserCredentials} UserCredentials
 */
```

designer/server/src/routes/forms/create.js
```diff
@@ -198,7 +200,13 @@ export default [
       // Submit new form metadata
       try {
         if (!result.error) {
```
colinrotherham, line 202: You'll have some type errors on the lines that follow

OpenID Connect (OIDC) doesn't guarantee `displayName` exists

So we'll need something like this with the current types:

```suggestion
        if (!result.error && credentials.user?.displayName) {
```

<br>

Or alternatively remove the question mark from the [`displayName?:` property](https://github.com/DEFRA/forms-designer/blob/2c556bb2a11eedae27d34f81fdd2aafd15b331c1/designer/server/src/types.ts#L60-L61)

Which only needs a simple check instead:

```suggestion
        if (!result.error && credentials.user) {
```

designer/server/src/routes/forms/create.js
```diff
@@ -198,7 +200,13 @@ export default [
...
-          await forms.create(result.value)
+          await forms.create({
+            metadata: result.value,
+            author: {
+              id: user.id,
+              displayName: user.displayName
+            }
+          })
```
colinrotherham, line 209: Once we've tightened the check for `credentials.user` we can destructure it without worrying that it might be undefined sometimes

This is using the 2x param suggestion for `forms.create(metadata, author)`

```suggestion
          const { id, displayName } = credentials.user
          await forms.create(result.value, { id, displayName })
```

designer/server/src/views/forms/library.njk
```diff
@@ -36,7 +36,7 @@
 
       {% set formUpdated %}
         <span class="app-display-until-desktop">Updated </span>{{ (form.draft.updatedAt if form.draft.updatedAt else "2024-04-09") | formatDate }}<br>
-        by {{ form.draft.updatedBy if form.draft.updatedBy else "Enrique Chase" }}
+        by {{ form.draft.updatedBy.displayName if form.draft.updatedBy else "Enrique Chase" }}
```
colinrotherham, line 39: Now these are getting longer, might be worth using the `default()` filter?

```suggestion
        by {{ form.draft.updatedBy.displayName | default("Enrique Chase", true) }}
```

designer/server/src/routes/forms/create.js
```diff
@@ -198,7 +200,14 @@ export default [
       // Submit new form metadata
       try {
         if (!result.error) {
-          await forms.create(result.value)
+          // @ts-expect-error - user is not undefined
```
colinrotherham, line 203: Sorry @davidjamesstone, can we remove this?

We can't guarantee that our code will always populate `credentials.user`

Let's use a tighter check from https://github.com/DEFRA/forms-designer/pull/148/files#r1595588500 instead

designer/server/src/lib/forms.js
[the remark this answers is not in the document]
davidjamesstone: Resolved with ca831a4382705fcff6c9409fe211b3343ed68d5a

designer/server/src/routes/forms/create.js
davidjamesstone: Added NonNullable<> ask discussed.
Added an exception with @ts-expect-error for the user is not undefined case.

designer/server/src/routes/forms/create.js
davidjamesstone: > Once we've tightened the check for credentials.user

We need to come back to look at this.
For now, I've added an exception with @ts-expect-error for the user is not undefined case.

designer/server/src/views/forms/library.njk
davidjamesstone: Resolved with ac0e839608f3a65b966d366edef24d3c7aa40402

designer/server/src/routes/forms/create.js
colinrotherham: Sorry @davidjamesstone

I'd prefer just tweaking the check then we can get this approved

We only use `@ts-expect-error` when we're purposely chucking incorrectly typed thing around

```suggestion
        if (!result.error && credentials.user) {
```

designer/server/src/routes/forms/create.js
```diff
@@ -13,6 +13,15 @@ import { buildErrorDetails } from '~/src/common/helpers/build-error-details.js'
 import * as forms from '~/src/lib/forms.js'
 import * as create from '~/src/models/forms/create.js'
 
+/**
+ * Gets the author from the auth credentials
+ * @param {import('@hapi/hapi').UserCredentials | undefined} user
```
colinrotherham, line 18: Mind keeping these in the footer?

We'll swap them all with `@import` blocks soon

Can use ~`user`~ `[user]` to handle potentially `undefined`

```suggestion
 * @param {UserCredentials} [user]
```

designer/server/src/routes/forms/create.js
```diff
@@ -13,6 +13,15 @@ import { buildErrorDetails } from '~/src/common/helpers/build-error-details.js'
...
+ */
+function getAuthor(user) {
+  // @ts-expect-error - user is not undefined
```
colinrotherham, line 21: Sorry Dave, we've said above that `user` might be `undefined`

Don't need this function if we do the extra check from https://github.com/DEFRA/forms-designer/pull/148#discussion_r1596733832

```suggestion
  if (!user) {
    return
  }

```

designer/server/src/routes/forms/create.js
```diff
@@ -183,7 +192,9 @@ export default [
     method: 'POST',
     path: '/create/team',
     async handler(request, h) {
-      const { payload, yar } = request
+      const { auth, payload, yar } = request
+      const { credentials } = auth
+      const { user } = credentials
```
colinrotherham, line 197: Just a reminder that this isn't correctly typed by Hapi

Sometimes `credentials` can be null so we'll need to check before destructuring

```suggestion
      const { user } = credentials
```

colinrotherham (approved):

alexluckett (approved):

alexluckett:
Great timing 😆

PR #149, opened 2024-05-13
Move slug formatting to `slugify` helper

colinrotherham:
Addresses code review feedback from https://github.com/DEFRA/forms-designer/pull/147/files#r1597036098

### DIFF
--- a/designer/server/src/routes/forms/create.js
+++ b/designer/server/src/routes/forms/create.js
@@ -1,7 +1,7 @@
 import {
   formMetadataInputSchema,
   organisationSchema,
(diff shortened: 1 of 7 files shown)


davidjamesstone (approved):

davidjamesstone:
LGTM

So we add this `slug` package into `FM` too?

colinrotherham:
Thanks @davidjamesstone

> So we add this `slug` package into `FM` too?

I've added it to `dependencies` so it'll be installed automatically

Just need to import `slugify` but I'll make another PR

PR #160, opened 2024-05-14
Add types and utilities for levels of authentication

colinrotherham:
This PR adds types for [`@hapi/bell`](https://hapi.dev/module/bell) `AuthArtifacts` immediately following authentication

With this we can do the following:

1. Add artifacts to `auth.credentials` such as **idToken**
2. Add helpers `hasAuthenticated()` (tokens are set) and `hasUser()` (user session available)
3. Add types to narrow `AuthCredentials` based on whether tokens or user checks passed

This means we can extract token properties `given_name`, `family_name` etc without extra checks

### DIFF
--- a/designer/server/src/common/helpers/auth/get-user-session.js
+++ b/designer/server/src/common/helpers/auth/get-user-session.js
@@ -6,22 +6,21 @@ import { token } from '@hapi/jwt'
  */
 export async function getUserSession(request, session) {
   const { auth, server } = request
+  const { credentials } = auth
 
   // Check for existing user
-  if (getUser(auth.credentials)) {
-    return auth.credentials
+  if (hasUser(credentials)) {
+    return credentials
   }
 
   // Prefer Session ID from cookie state
   let sessionId = session?.sessionId
 
   // Fall back to OpenID Connect (OIDC) claim
-  if (!sessionId) {
-    const claims = getUserClaims(auth.credentials)
+  if (!sessionId && hasAuthenticated(credentials)) {
+    const { accessToken } = getUserClaims(credentials)
 
-    if (claims?.sub) {
-      sessionId = claims.sub
-    }
+    sessionId = accessToken.sub
   }
 
   // Retrieve user session from Redis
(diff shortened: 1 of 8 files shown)


designer/server/src/common/helpers/auth/get-user-session.js
```diff
@@ -32,28 +31,37 @@ export async function getUserSession(request, session) {
...
+  if (!decoded.payload) {
+    throw new Error('Failed to decode access token')
+  }
+
   return decoded.payload
 }
 
 /**
  * @param {AuthCredentials | null} [credentials]
+ * @returns {credentials is AuthSignedIn}
```
colinrotherham, line 61: Should `hasUser()` return true, we narrow `AuthCredentials` to `AuthSignedIn`

This stops us needing `credentials?.user?.id` or other type checks

designer/server/src/common/helpers/auth/get-user-session.js
```diff
@@ -32,28 +31,37 @@ export async function getUserSession(request, session) {
 
 /**
  * @param {AuthCredentials | null} [credentials]
+ * @returns {credentials is AuthWithTokens}
```
colinrotherham, line 34: Should `hasAuthenticated()` return true, we narrow `AuthCredentials` to `AuthTokens`

This still leaves `user` as optional but prevents unnecessary token checks

designer/server/src/common/helpers/auth/user-session.js
```diff
@@ -25,25 +27,33 @@ export function createUser(credentials) {
...
 export async function createUserSession(request) {
   const { auth, server } = request
+  const { artifacts, credentials } = auth
 
-  // Optionally create user object (e.g. signed in token but no session)
-  const user = !getUser(auth.credentials)
-    ? createUser(auth.credentials)
-    : auth.credentials.user
+  // Patch missing properties using Bell artifacts
+  credentials.idToken = artifacts.id_token
```
colinrotherham, line 37: I've added types for `AuthArtifacts` so we can pull `artifacts.id_token` from Bell

alexluckett (approved):

PR #161, opened 2024-05-14
Decode ID token via `getUserClaims()`

colinrotherham:
This PR updates our `getUserClaims()` helper to add the decoded ID token


alexluckett (approved):

designer/server/src/common/helpers/auth/get-user-session.js
```diff
@@ -72,7 +86,7 @@ export function hasUser(credentials) {
  */
 
 /**
- * @typedef {Pick<AuthCredentials, 'token' | 'idToken' | 'refreshToken'>} Tokens - Known tokens
+ * @typedef {Pick<AuthCredentials, 'token' | 'idToken'>} Tokens - Known tokens
```
colinrotherham, line 89: I've temporarily removed `refreshToken` from the list we decode

It's still available via `credentials.refreshToken` as a string

PR #167, opened 2024-05-15
Add navigation entries for form overview and editor pages

colinrotherham:
This PR updates the `navigation` context for form overview and editor pages

I've also updated the `buildEntry()` helper to work without a Hapi request object

<br>

Navigation entries

### DIFF
--- a/designer/server/src/common/templates/partials/navigation.njk
+++ b/designer/server/src/common/templates/partials/navigation.njk
@@ -3,7 +3,7 @@
 
 {{ appServiceHeader({
   organisationName: "Defra",
-  productName: "Forms designer",
+  productName: "Forms Designer",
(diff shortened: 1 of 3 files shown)


davidjamesstone (approved):

designer/server/src/common/templates/partials/navigation.njk
```diff
@@ -3,7 +3,7 @@
 
 {{ appServiceHeader({
   organisationName: "Defra",
-  productName: "Forms designer",
+  productName: "Forms Designer",
```
colinrotherham, line 6: Just highlighting this as product names are typically title case (unlike service names)

See guidance on [**Nunjucks macro options** `productName`](https://design-system.service.gov.uk/components/header/#options-default-1)

PR #172, opened 2024-05-17
Set up "two-thirds" grid columns for headings etc

colinrotherham:
Using changes from https://github.com/DEFRA/forms-designer/pull/169 and https://github.com/DEFRA/forms-designer/pull/171 this PR add the following:

1. Changes the "Organisation details" heading to an `<h2>`
2. Sets up the correct "two-thirds" grid columns widths where necessary
3. Adds page heading params to routes with models

## Before

Before

## After

After

### DIFF
--- a/designer/server/src/models/account/auth.js
+++ b/designer/server/src/models/account/auth.js
@@ -1,5 +1,11 @@
 export function signedOutViewModel() {
+  const pageTitle = 'You have signed out'
+
   return {
-    pageTitle: 'You have signed out'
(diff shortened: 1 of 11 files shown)


colinrotherham:
Another screenshot

I've ensured the nested grid column for the heading switches to "full" on desktop

This is because it's already within a "two-thirds" column at this breakpoint

After (Desktop)

davidjamesstone (approved):

PR #174, opened 2024-05-17
Add author to the update definition route

davidjamesstone:


designer/server/src/routes/forms/api.js
```diff
@@ -33,16 +33,17 @@ export default [
         parse: true
       },
       async handler(request, h) {
-        const { id } = request.params
+        const { auth, params, payload } = request
+        const { id } = params
+        const author = forms.getAuthor(auth.credentials)
```
colinrotherham, line 38: Do you mind me saying it feels a bit overly-destructured?

E.g. I quite like seeing `params.id` to know where it comes from

```suggestion
        const author = forms.getAuthor(auth.credentials)
```

designer/server/src/routes/forms/api.js
```diff
@@ -33,16 +33,17 @@ export default [
...
-              abortEarly: false
-            }
-          )
+          const result = formDefinitionSchema.validate(payload, {
+            abortEarly: false
+          })
 
-          if (error) {
+          if (result.error) {
+            const error = result.error
```
colinrotherham, line 46: Similar with this, feels like it didn't need another const?

```suggestion
```

<br>

Using `result.error` is a bit more obvious

```patch
-             [error, request.payload]
+             [result.error, request.payload]
```

designer/server/src/routes/forms/api.js
```diff
@@ -53,7 +54,10 @@ export default [
             )
           }
 
-          await forms.updateDraftFormDefinition(id, value)
+          const value = result.value
```
colinrotherham, line 54: Maybe we don't need `const value` here either?

But it might be a nice way to show the schema value is a form `definition`?

```suggestion
          const { value: definition } = result
```

designer/server/src/routes/forms/api.js
```diff
@@ -33,16 +33,17 @@ export default [
...
+        const author = forms.getAuthor(auth.credentials)
+
         try {
-          const { value, error } = formDefinitionSchema.validate(
-            request.payload,
-            {
-              abortEarly: false
-            }
-          )
+          const result = formDefinitionSchema.validate(payload, {
```
colinrotherham, line 41: Thanks for deferring the destructuring to fix that ESLint warning 🙌

designer/server/src/routes/forms/api.js
davidjamesstone: I have a rule, if I reference an object property more than once, I'll create a variable for it.
I find that it keeps things more readable! Personally, I don't like `params.this` and `request.that` :see_no_evil: 

Also, there's a cost to the JS interpreter each time you look up an key on an object. While these days it has probably been optimised out in the JIT compiler, old habits and all.

designer/server/src/routes/forms/api.js
davidjamesstone: All the handlers do is basically destructring and passing along data.
I find this reads better.

Also, it "scales" better IMO:

E.g.Imagine a route starts out like this:

```
const { params, payload } = request
const { id } = params
const author = forms.getAuthor(auth.credentials)
```

Then sometime later, a new parameter gets added, and the whole route gets authenticated. It's easy to simply extend the destructring:

```
const { auth, params, payload } = request
const { id, slug } = params
```

designer/server/src/routes/forms/api.js
davidjamesstone: Kinda wishing I hadn't bothered! :rofl: - it takes longer to then discuss code styles!

colinrotherham: Nah it's all worth it and this feedback is good, otherwise we'll each do things 3x different ways!

Good to leave things in order for the next person

PR #184, opened 2024-05-21
Various form editor React UI font size tweaks and fixes

colinrotherham:
This PR avoids the deprecated font `$size: 14` and fixes some spacing issues

I'll need to use ~`govuk-font()`~ `govuk-font-size()` selectively once we update to GOV.UK Frontend v5

The close button is now also a `<button>`


designer/client/src/components/ComponentCreate/__snapshots__/ComponentCreateList.test.tsx.snap
```diff
@@ -31,79 +31,103 @@ exports[`ComponentCreateList should match snapshot 1`] = `
           class="govuk-list"
         >
           <li>
-            <a
```
alexluckett, line 34: isn't this file a test artefact, rather than something that needs committing? I realise you've only modified it not introduced it though :-)

colinrotherham: Yeah it is, but you're supposed to commit them 😣

I'd prefer if we removed it at some point, but it's generated by one test:

```js
expect(asFragment()).toMatchSnapshot()
```

Whenever you change the source markup, Jest complains that the snapshot needs updating via:

```console
npm test -- -u
```

alexluckett: My apologies! So they are.

alexluckett (approved):

PR #186, opened 2024-05-22
Fix code editor crash affecting various components

colinrotherham:
This PR fixes a crash in the [code editor](https://www.npmjs.com/package/react-simple-code-editor) due to a synthetic default export

This was complicated by the way we import and bundle our code

* Editor component needed the `*.jsx` extension
* TypeScript needed the option [`"moduleResolution": "bundler"`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler)
* Babel [`modules` transform](https://babeljs.io/docs/babel-preset-env#modules) now turned off

I've also picked up styling changes from:

* https://github.com/DEFRA/forms-designer/pull/185


alexluckett (commented):
Excellent spot :-) 

These 3 test failures are new though. they're currently passing on main.

colinrotherham:
Thanks @alexluckett

>These 3 test failures are new though. they're currently passing on main.

Yeah bit annoying that! Thankfully it's only config tweaks so I'm taking a look! 👀

Currently if you fix one you break the other:

## Works in Jest (but not browsers)
Babel via Jest is handling "synthetic default exports" from ES modules compiled to CommonJS

```jsx
import SimpleEditor from 'react-simple-code-editor'

<SimpleEditor />
```

## Works in browsers (but not Jest)
Babel via webpack is not handling "synthetic default exports" from ES modules compiled to CommonJS

```jsx
import SimpleEditor from 'react-simple-code-editor'

<SimpleEditor.default />
```

alexluckett (approved):

PR #196, opened 2024-05-24
Fail on startup if env vars aren't defined

alexluckett:
Too many startup errors are being swallowed by missing env vars. This makes them required so the error is explicit and caught on startup, rather than requiring checks further in the codebase.

Any default env vars required can be set in https://github.com/DEFRA/cdp-app-config/.

These defaults are added back in https://github.com/DEFRA/cdp-app-config/pull/418.


designer/server/jest.setup.cjs
```diff
@@ -1,2 +1,15 @@
 process.env.SESSION_COOKIE_PASSWORD = 'test-env-session-cookie-password'
 process.env.ROLE_EDITOR_GROUP_ID = 'valid-test-group'
+process.env.MANAGER_URL = 'http://localhost'
+process.env.PREVIEW_URL = 'http://localhost'
+process.env.SESSION_TTL = '86400000'
+process.env.SESSION_COOKIE_TTL = '2419200000'
```
colinrotherham, line 6: Ah I think you've swapped these around by accident?

Do you mind keeping the time code so it's always expressive and human readable?

```suggestion
process.env.SESSION_TTL = Duration.fromObject({ days: 1 })
  .as('milliseconds')
  .toString()
process.env.SESSION_COOKIE_TTL = Duration.fromObject({ minutes: 30 })
  .as('milliseconds')
  .toString()
```

designer/server/jest.setup.cjs
```diff
@@ -1,2 +1,15 @@
 process.env.SESSION_COOKIE_PASSWORD = 'test-env-session-cookie-password'
 process.env.ROLE_EDITOR_GROUP_ID = 'valid-test-group'
+process.env.MANAGER_URL = 'http://localhost'
```
colinrotherham, line 3: Can we keep the port and trailing slash?

```suggestion
process.env.MANAGER_URL = 'http://localhost:3001/'
```

designer/server/jest.setup.cjs
```diff
@@ -1,2 +1,15 @@
...
+process.env.PREVIEW_URL = 'http://localhost'
```
colinrotherham, line 4:
```suggestion
process.env.PREVIEW_URL = 'http://localhost:3009'
```

colinrotherham (commented):
Looks good, can you double check those TTL times?

If you can ordering the envs A-Z it helps with readability too

Thanks @alexluckett

designer/server/jest.setup.cjs
alexluckett: 🤦  I'm thinking this is a just a key/value properties file, of course it's Javascript! Yes, absolutely we should leave that in.

colinrotherham (approved):

PR #207, opened 2024-05-30
Restore previous URL path after auth callback

colinrotherham:
This PR ensures the previous URL path is restored after your session expires

We'll only be handling the URL path so I've renamed the yar key from `referrer` (full URL) → `redirectTo`


colinrotherham:
Forgot this was the base branch for https://github.com/DEFRA/forms-designer/pull/208

Will merge this one too

PR #232, opened 2024-06-05
Remove redis reconnect

alexluckett:
`redisClient.on('close', ....)` needs to be removed as it prevents the connection retry logic kicking in that's built into the library by default.


designer/server/src/common/helpers/redis-client.js
```diff
@@ -50,15 +50,14 @@ function buildRedisClient() {
...
-
   redisClient.on('connect', () => {
     logger.info('Connected to Redis server')
   })
 
   redisClient.on('close', () => {
-    logger.info('Redis connection closed attempting reconnect')
-    redisClient.connect()
+    logger.warn(
+      'Redis connection closed attempting reconnect with default behavior'
```
colinrotherham, line 59: Not a blocker, but should probably use the `reconnecting` event to confirm this behaviour

* `connect` Connected to Redis server
* `close` Connection to Redis server closed
* `reconnecting` Reconnecting to Redis server
* `error` Error connecting to Redis server

We should also consider `enableReadyCheck: true` and wait for the `ready` event before server start

```suggestion
      'Connection to Redis server closed'
```

alexluckett: I'll take this up as follow-up action separately.

PR #234, opened 2024-06-07
Add outputEmail to formdefinition schema

davidjamesstone:


alexluckett (approved):

model/src/form/form-definition/index.ts
```diff
@@ -260,7 +260,10 @@ export const formDefinitionSchema = Joi.object<FormDefinition>()
     outputs: Joi.array<Output>().items(outputSchema),
     skipSummary: Joi.boolean().optional().default(false),
     phaseBanner: phaseBannerSchema,
-    specialPages: specialPagesSchema.optional()
+    specialPages: specialPagesSchema.optional(),
+    outputEmail: Joi.string()
+      .email({ tlds: { allow: ['uk'] } })
```
colinrotherham, line 265: Noticed in `teamEmailSchema` (and other user-input fields) we have `trim()`?

```suggestion
      .email({ tlds: { allow: ['uk'] } })
      .trim()
```

colinrotherham (approved):

model/src/form/form-definition/index.ts
davidjamesstone: Resolved with 37397a4b94f710f9bea53c882763185499458492

PR #235, opened 2024-06-07
Prefer enums for component types + fixes

colinrotherham:
This PR moves all component types to use the existing enums

When removing components this will let the compiler flag any missed code to avoid bugs

Some fixes are also included:

1. Fix Joi validation for missing `subType` component property
2. Add missing `schema` property on component create
3. Add missing `WebsiteField` component type enum

With the following code removed:

1. Remove unused `FormConfiguration()` class
2. Remove unused `ConditionalComponentTypes`


alexluckett (approved):
Love this!

PR #258, opened 2024-06-13
Fix editor API logging

colinrotherham:
This PR fixes a Node.js error trying to log calls to `/api/log`

>SyntaxError: The string did not match the expected pattern.

```js
request.server.log(request.payload.toString())
```

I've also included changes to:

1. Log errors as the first parameter
2. Ensure error objects are serialised by `pino`
3. Add env `REACT_LOG_LEVEL` to tests


davidjamesstone:
I prefer `err` over `error` and `cause` in a try/catch personally,
let's agree a standard/enforce something with eslint across the repos?

davidjamesstone (approved):

PR #259, opened 2024-06-13
Fix bug affecting `hasValidated` state

colinrotherham:
This PR fixes a bug where `hasValidated` was set for all component actions

I've switched to the original names as non-option `EDIT_*` and `SET_*` events would be great for auditing

```patch
  import {
-   Actions
+   Fields,
+   Meta,
+   Options,
+   Schema
  } from '~/src/reducers/component/types.js'
```

### DIFF
--- a/designer/client/src/reducers/component/componentReducer.tsx
+++ b/designer/client/src/reducers/component/componentReducer.tsx
@@ -43,16 +43,6 @@ export const ComponentContext = createContext<{
 })
 
 /**
- * Action types combined
- */
-export const actionTypes = {
-  ...Meta,
-  ...Schema,
-  ...Fields,
-  ...Options
-}
-
-/**
  * Reducers mapped by action type
  */
 const reducerByActionType = [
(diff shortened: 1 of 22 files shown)


designer/client/src/reducers/component/componentReducer.tsx
```diff
@@ -42,45 +43,41 @@ export const ComponentContext = createContext<{
...
 
 /**
- * when an {@link Actions} is passed to getSubReducer, it will return the associated reducer defined in {@link ActionsReducerCollection}
+ * when an enum is passed to getSubReducer, it will return the associated reducer defined in {@link reducerByActionType}
  */
-export function getSubReducer(type) {
-  return ActionsReducerCollection.find((a) => valueIsInEnum(type, a[0]))?.[1]
-}
-
-const isNotValidate = (type): type is Meta.VALIDATE => {
```
colinrotherham, line 65: The function `isNotValidate()` allowed all actions through

davidjamesstone (approved):

PR #260, opened 2024-06-13
Request fresh form definition after save

colinrotherham:
This PR ensures form definitions are re-fetched after each save rather than trusting the React state

I've also replaced our `new DesignerApi()` class with a form library in a familiar style but using [**Fetch** API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)

```js
import * as form from '~/src/lib/form.js'

// Get definition
await form.get(id)

// Save changes
await form.save(id, definition)

// Log events
await form.log(error, ...messages)
```


designer/client/src/Designer.tsx
```diff
@@ -38,32 +34,36 @@ export class Designer extends Component<Props, State> {
...
+    })
+
+    return definition
   }
 
-  updatePageContext = (page) => {
-    this.setState({ page })
+  save = async (definition: FormDefinition) => {
+    await form.save(this.id, definition)
+    return this.get()
```
colinrotherham, line 59: The save response returns the `{ ok: true }` pattern rather than the definition

It also causes `response.json()` parsing errors

davidjamesstone (approved):

PR #265, opened 2024-06-14
Fix textarea "rows" option

colinrotherham:
This PR fixes the "rows" option when editing a Textarea component (multi-line text)

# Rows option

The rows option was a single `<input>` without a label, resting on top of **Autocomplete**

## Before

Rows option before

## After

Rows option after

# Additional settings

These settings were missing a wrapper so fixed them at the same time

## Before

Additional settings before

## After

Additional settings after

### DIFF
--- a/designer/client/src/components/FieldEditors/TextFieldEdit.tsx
+++ b/designer/client/src/components/FieldEditors/TextFieldEdit.tsx
@@ -27,144 +27,146 @@ export function TextFieldEdit({ children, context = ComponentContext }: Props) {
         </span>
       </summary>
 
-      <div className="govuk-form-group">
-        <label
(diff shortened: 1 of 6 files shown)


designer/client/src/components/FieldEditors/TextFieldEdit.tsx
```diff
@@ -131,6 +131,8 @@ export function TextFieldEdit({ children, context = ComponentContext }: Props) {
         />
       </div>
 
+      {children}
```
colinrotherham: Moved this up so **Rows** is grouped with the other min/max options rather than after **Regex**

alexluckett (approved):

PR #268, opened 2024-06-14
Fix lots of editor page + link bugs

colinrotherham:


designer/client/src/data/page/deleteLink.ts
```diff
@@ -0,0 +1,28 @@
...
+import { type Path } from '~/src/data/types.js'
+
+export function deleteLink(
+  data: FormDefinition,
+  from: Path,
+  to: Path
+): FormDefinition {
+  const [fromPage] = findPage(data, from)
+
+  findPage(data, to)
```
davidjamesstone, line 13: Is it worth commenting that it's being called for side-effects (to throw?).
Or instead using the return value from it in the line following:

```
const [toPage] = findPage(data, to)

const toLinkIndex = fromPage.next?.findIndex((n) => n.path === toPage.path) ?? -1
```

It looks a bit lost otherwise.

davidjamesstone (approved):

designer/client/src/data/page/deleteLink.ts
colinrotherham: Yeah it does. A few of the existing utilities did the same so I kept the status quo

PR #269, opened 2024-06-14
Fix or remove `startPage` on definition save (+ upload)

colinrotherham:
This PR ensures the editor fixes or removes incorrect `startPage` paths

Two new utilities are now exported from `@defra/forms-model`:

1. `findStartPage()`
2. `updateStartPage()`

With the lookup code moved from:

* https://github.com/DEFRA/forms-manager/pull/185

Credit to @alexluckett for the original PR (description below)

---

Previously, when we renamed a start page in the form editor, the `startPage` would be updated too. ~We've removed this, so we need an alternative way of setting the `startPage`.~

The original method was flawed anyway, as it only handled page renames, but it didn't account for users who create entirely new start pages. Imagine I have an existing branch on a form (with a start page), but I create a new branch with a new start page and then delete my old branch, the start page never have been updated.

As an example, deleting the original "First page" and replacing it with "new start page":

image

This way, when a user saves a form, the start page is automatically detected. It supports all scenarios, e.g.:

- Fixes: Manually updating the JSON and putting in an incorrect `startPage` entry
- Fixes: Changing the start page, ~leaving~ removing the old `startPage` ~in place~ to prevent 404s
- Fixes: Renaming the start page

Longer term we might want to add UI functionality for the start page to be set. Or better yet, keep the autodetection but add some proper error handling in place to guide the users to fix their mistakes (a form can't have multiple start pages).

### DIFF
--- a/designer/client/src/Designer.tsx
+++ b/designer/client/src/Designer.tsx
@@ -1,11 +1,14 @@
-import { type FormDefinition, type FormMetadata } from '@defra/forms-model'
+import {
+  updateStartPage,
+  type FormDefinition,
+  type FormMetadata
+} from '@defra/forms-model'
 import React, { Component } from 'react'
 
 import { Menu } from '~/src/components/Menu/Menu.jsx'
 import { Visualisation } from '~/src/components/Visualisation/Visualisation.jsx'
 import { DataContext } from '~/src/context/DataContext.js'
 import { FlyoutContext } from '~/src/context/FlyoutContext.js'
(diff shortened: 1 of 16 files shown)


colinrotherham:
The changes in this PR led to [fixes in commit `3a417e6`](https://github.com/DEFRA/forms-designer/pull/269/commits/3a417e6851225350b9f8836d591d2066e7bcc8aa) where test fixtures needed to be linear form journeys

alexluckett (approved):

PR #324, opened 2024-07-09
Add form metadata to info tabs, rename to "Form overview"

colinrotherham:
Rather than remove "Summary" this renamed it to "Form overview" and drops it into the sub navigation

## Navigation changes
* The button "Summary" is now for the actual check answers summary
* The form summary tabs have moved into the sub navigation as "Form overview"

Navigation menu

## Form overview tabs
Since this was a useful feature, I've added the "Form metadata" tab and turned on syntax highlighting

New form overview tab layout

### DIFF
--- a/designer/client/src/Designer.tsx
+++ b/designer/client/src/Designer.tsx
@@ -20,16 +20,18 @@ interface Props {
 interface State {
   flyoutCount: number
   data: FormDefinition
+  meta: FormMetadata
 }
 
 export class Designer extends Component<Props, State> {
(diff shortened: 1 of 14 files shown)


alexluckett (commented):
You've got a unit test failure on this but the change is looking good otherwise!

colinrotherham:
>You've got a unit test failure on this but the change is looking good otherwise!

@alexluckett Cheers, forgot to rename the menu item in the test too

All sorted

alexluckett (approved):

PR #328, opened 2024-07-12
Feature 398639: Change lead organisation for a form

alexluckett:
- Removes the boolean success notification flags used on the form overview page, replaces with a global success notification state attribute
- Introduces an edit page for the form's lead organisation, re-using the existing creation page view and model (with an override for the back link to the user to the form overview page)
- Upon success, display the notification to the user

⚠️  Depends on https://github.com/DEFRA/forms-manager/pull/223. Do not merge until this dependency is merged into forms-manager.

image
image
image


davidjamesstone (approved):

colinrotherham (changes requested):
I'm still seeing lots of `designer/server/*` ESLint issues

Might be a setup issue, have a look at the checks?

alexluckett:
Weird, these aren't coming through locally. Looking into it...

alexluckett:
Fixed. As discussed on Slack, this was caused by the pipeline not building the model first. Fixed in https://github.com/DEFRA/forms-designer/pull/334.

designer/server/src/lib/forms.js
```diff
@@ -125,6 +125,28 @@ export async function createDraft(id, token) {
...
 
+/**
+ * Updates a metadata object.
+ * @param {string} id
+ * @param {Partial<FormMetadataInput>} metadata
+ * @param {string} token - auth token
+ * @returns
+ */
+export async function updateMetadata(id, metadata, token) {
+  const patchJsonByType = /** @type {typeof patchJson<FormMetadataInput>} */ (
```
colinrotherham, line 136: Ahh! Mind checking this? I think we return `{ id, slug, status }`?

We definitely set the wrong `body` type already in places and might be spreading it further here

designer/server/src/models/forms/edit.js
```diff
@@ -0,0 +1,21 @@
+import * as create from '~/src/models/forms/create.js'
+
+/**
+ * @param {string} slug
+ * @param {Partial<FormMetadataInput>} [metadata]
```
colinrotherham, line 5: Looks like we call `organisationViewModel()` after a full metadata response:

```js
```

Presume the slug via `metadata.slug` works with ~`FormMetadataInput`~ `FormMetadata`

```suggestion
 * @param {FormMetadata} metadata
```

designer/server/src/models/forms/edit.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {ValidationFailure} [validation]
+ */
+export function organisationViewModel(slug, metadata, validation) {
+  return {
+    ...create.organisationViewModel(metadata, validation),
+    backLink: {
+      href: `/library/${slug}`
```
colinrotherham, line 12: We set the text usually, that alright?

```suggestion
      text: 'Back to form overview',
      href: `/library/${metadata.slug}`
```

designer/server/src/models/forms/library.js
```diff
@@ -20,24 +20,14 @@ export async function listViewModel(token) {
 
 /**
  * @param {FormMetadata} metadata
- * @param {boolean} [displayCreateLiveSuccess] - indicating if the form was successfully promoted to live
- * @param {boolean} [displayCreateDraftSuccess] - indicating if a draft form was successfully created
+ * @param {string} [notification] - success notification to display
```
colinrotherham, line 23: Cheers for combining these, remember we talked about it a while back 🙌

designer/server/src/routes/forms/create.js
```diff
@@ -34,13 +41,31 @@ const schema = Joi.object().keys({
   })
 })
 
+/**
+ * @type {FailAction}
+ */
+export const displayJoiFailures = (request, h, error) => {
```
colinrotherham, line 47: We tend to export functions so the `displayJoiFailures` name appears in stack traces

Rather than the anonymous array function

```suggestion
export function displayJoiFailures(request, h, error) {
```

designer/server/src/routes/forms/create.js
```diff
@@ -49,16 +74,15 @@ export default [
       yar.clear(sessionNames.validationFailure)
 
       // Redirect to first step
-      return h.redirect('/create/title').temporary()
+      return h.redirect(ROUTE_PATH_CREATE_TITLE).temporary()
     }
   }),
-
```
colinrotherham, line 55: Mind adding these spaces back in?

It was getting a bit cramped so I spaced out the JSDoc comments

Can make it an ESLint rule if it's something we can auto-format next time

designer/server/src/routes/forms/create.js
colinrotherham: Could rename it to `redirectToStepWithErrors()` and include the "which has the error?" logic from the final fail action if that's helpful? These checks etc:

```js
// Optionally redirect to errors on previous steps
if ('title' in formErrors) {
  redirectTo = '/create/title'
} else if ('organisation' in formErrors) {
  redirectTo = '/create/organisation'
}
```

We only put it at the end to avoid duplication but don't need to now

designer/server/src/models/forms/edit.js
alexluckett: Excellent spot! Not sure why I originally started down that route. Fixed.

designer/server/src/models/forms/edit.js
alexluckett: Done

designer/server/src/routes/forms/create.js
alexluckett: Done, consolidated.

designer/server/src/routes/forms/create.js
alexluckett: Done

designer/server/src/lib/forms.js
alexluckett: Good spot, thanks. I wasn't even using the return value which is why I didn't spot it before.

I've added a type to the model and set it here. Fixed.

colinrotherham: Great, thanks for this

Can we change the name `FormUpdateResponse` though?

We've got the same response via **POST** when creating forms too

designer/server/src/routes/forms/create.js
```diff
@@ -40,7 +47,7 @@ export default [
    */
   ({
     method: 'GET',
-    path: '/create',
+    path: ROTUE_PATH_CREATE,
```
colinrotherham, line 50:
```suggestion
    path: ROUTE_PATH_CREATE,
```

designer/server/src/routes/forms/create.js
```diff
@@ -316,3 +320,7 @@ function redirectToTitleWithErrors(request, h) {
  * @typedef {import('@hapi/hapi').ResponseToolkit<any>} ResponseToolkit
  * @typedef {Request<{ Payload: any }>} RequestWithPayload
  */
+
+/**
+ * @typedef {import('@hapi/hapi').RouteOptionsValidate['failAction']} FailAction
```
colinrotherham, line 325: The `failAction` property is already typed if you want to use `Lifecycle.FailAction` or `FailAction`?

```suggestion
 * @typedef {import('@hapi/hapi').Lifecycle} Lifecycle
```

```suggestion
 * @typedef {import('@hapi/hapi').Lifecycle.FailAction} FailAction
```

designer/server/src/routes/forms/create.js
```diff
@@ -295,7 +250,56 @@ function redirectToTitleWithErrors(request, h) {
...
+}
+
+/**
+ * @param {Request} request
+ * @param {ResponseToolkit} h
+ * @param {Error} [error]
+ */
+export function redirectToStepWithErrors(request, h, error) {
+  return redirectWithErrors(request, h, error, true)
+}
```
colinrotherham, line 266: Do we need this wrapper function?

It's the only place we call `redirectWithErrors()` and the `redirectToPreviousStep` param never changes

designer/server/src/routes/forms/create.js
alexluckett: 🙈

designer/server/src/routes/forms/create.js
alexluckett: It's used from `edit.js` as well.

In `create.js`, redirectWithErrors is called with `true`.
In `edit.js`, `redirectWithErrors` is called with `false`.

designer/server/src/routes/forms/create.js
alexluckett: Thanks! I've removed it, I actually I wasn't even using this type, it was leftover code from a previous iteration.

designer/server/src/routes/forms/create.js
alexluckett: Done

designer/server/src/lib/forms.js
alexluckett: Done

PR #340, opened 2024-07-16
Add createdBy, createdAt, updatedBy, updatedAt to form metadata as top-level attributes

alexluckett:


model/src/form/form-metadata/types.ts
```diff
@@ -84,12 +84,39 @@ export interface FormMetadata {
...
   FormMetadata,
-  'id' | 'slug' | 'draft' | 'live'
+  | 'id'
+  | 'slug'
+  | 'draft'
+  | 'live'
+  | 'createdBy'
+  | 'createdAt'
+  | 'updatedBy'
+  | 'updatedAt'
```
colinrotherham, line 121: If you find this a bit much, you can use `Pick<FormMetadata, 'one' | 'two'>` instead?

i.e. Pick what's allowed versus remembering to exclude what's not

alexluckett: That works too! Updated.

colinrotherham (approved):

davidjamesstone (approved):

alexluckett:
Dave's happy too.

PR #342, opened 2024-07-19
392202 - Reorder components

davidjamesstone:
WIP.

Dan S is happy with this approach but there's some accessibilty imrpovements with the show/hide to be made.


alexluckett:
Loving this!

alexluckett (approved):

alexluckett:
As discussed on Slack, I think we need to do a job at visually letting the user know what component is in what place as a group of similar looking components will appear to not move, e.g. in the below form.

image

I'll flag this to Dan on Monday, happy for it to go through in meantime.

PR #356, opened 2024-07-23
#398639 Use the form's last updated in the library page

alexluckett:
Since https://github.com/DEFRA/forms-designer/pull/340 we now have top-level form updated by/at values. This PR uses them on the library page, rather than the value from the draft/live.

This is being brought in under ticket #398639, [based on our Slack conversation](https://defra-digital-team.slack.com/archives/C071FRPN9TK/p1721138889548189).

### DIFF
--- a/designer/server/src/views/forms/library.njk
+++ b/designer/server/src/views/forms/library.njk
@@ -33,11 +33,9 @@
         </a>
       {% endset %}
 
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):
Feels a little patchy to show "Unknown" when we have the info

But approving this knowing we'll fill the gaps over time

davidjamesstone (approved):

PR #418, opened 2024-08-27
Update link helpers to clone before modifying data

colinrotherham:
To unblock https://github.com/DEFRA/forms-designer/pull/419 this PR avoids `{ ...data }` props copied by reference in link helpers

```js
return {
    ...data,
    pages: data.pages.map(…)
}
```

It follows on from:

* https://github.com/DEFRA/forms-designer/pull/411

### DIFF
--- a/designer/client/src/PageEdit.tsx
+++ b/designer/client/src/PageEdit.tsx
@@ -94,8 +94,9 @@ export class PageEdit extends Component<Props, State> {
     pageEdit.section = section?.name
 
     if (payload.path !== page.path) {
-      copy = updateLinksTo(copy, page.path, payload.path)
-      pageEdit.path = payload.path
+      copy = updateLinksTo(copy, pageEdit, {
+        path: payload.path
+      })
     }
 
     try {
--- a/designer/client/src/PageEdit.tsx
+++ b/designer/client/src/PageEdit.tsx
@@ -15,7 +15,9 @@ import { logger } from '~/src/common/helpers/logging/logger.js'
 import { Flyout } from '~/src/components/Flyout/Flyout.jsx'
 import { RenderInPortal } from '~/src/components/RenderInPortal/RenderInPortal.jsx'
 import { DataContext } from '~/src/context/DataContext.js'
+import { deleteLink } from '~/src/data/page/deleteLink.js'
 import { findPage } from '~/src/data/page/findPage.js'
+import { hasNext } from '~/src/data/page/hasNext.js'
 import { updateLinksTo } from '~/src/data/page/updateLinksTo.js'
 import { findSection } from '~/src/data/section/findSection.js'
 import { controllerNameFromPath } from '~/src/helpers.js'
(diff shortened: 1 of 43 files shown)


designer/client/src/PageEdit.tsx
```diff
@@ -84,19 +86,18 @@ export class PageEdit extends Component<Props, State> {
...
     }
 
-    copyPage.title = payload.title
-    copyPage.controller = controller
-    copyPage.section = section?.name
-
-    copy.pages[copyIndex] = copyPage
+    pageEdit.title = payload.title
+    pageEdit.controller = controller
+    pageEdit.section = section?.name
```
colinrotherham, line 100: I'm going to move these lines above `updateLinksTo()`

This is because `pageEdit` becomes stale when `definition` is cloned

designer/client/src/PageEdit.tsx
```diff
@@ -141,22 +142,24 @@ export class PageEdit extends Component<Props, State> {
...
     // Remove all links to the page
-    copy.pages.forEach((p, index) => {
-      if (index !== copyPageIdx && Array.isArray(p.next)) {
-        for (let i = p.next.length - 1; i >= 0; i--) {
-          const next = p.next[i]
-          if (next.path === page.path) {
-            p.next.splice(i, 1)
-          }
-        }
+    for (const pageFrom of pages.filter(hasNext)) {
```
colinrotherham, line 152: Just to avoid confusion, I'm going to change `pages` → `copy.pages`

Important to remember that `deleteLink` clones `copy` so `{ pages }` is from another reference

```suggestion
    for (const pageFrom of copy.pages.filter(hasNext)) {
```

designer/client/src/PageEdit.tsx
```diff
@@ -141,22 +142,24 @@ export class PageEdit extends Component<Props, State> {
...
+      // Remove link
+      if (next.some(({ path }) => path === pageRemove.path)) {
+        copy = deleteLink(copy, pageFrom, pageRemove)
       }
-    })
+    }
 
-    copy.pages.splice(copyPageIdx, 1)
+    // Remove page
+    pages.splice(pageIndex, 1)
```
colinrotherham, line 162: Same here. We need to make sure we remove the page from the latest reference

```suggestion
    // Remove page
    copy.pages.splice(pageIndex, 1)
```

designer/client/src/data/page/updateLinksTo.ts
```diff
@@ -1,24 +1,43 @@
...
+
+  // Copy form definition
+  const definition = structuredClone(data)
+  const { pages } = definition
+
+  // Check for outdated pages
+  for (const pageEdit of pages) {
+    if (!hasNext(pageEdit)) {
+      continue
+    }
```
colinrotherham, line 24: I need to move this `hasNext()` to after the page path is updated

Otherwise pages without links (e.g. the summary page) won't see the update

PR #449, opened 2024-09-06
Edit privacy notice url in form metadata

davidjamesstone:
image
image
image
image

### DIFF
--- a/designer/server/src/routes/forms/privacy-notice.js
+++ b/designer/server/src/routes/forms/privacy-notice.js
@@ -1,0 +1,120 @@
+import { privacyNoticeUrlSchema } from '@defra/forms-model'
+import { StatusCodes } from 'http-status-codes'
+import Joi from 'joi'
+
+import * as scopes from '~/src/common/constants/scopes.js'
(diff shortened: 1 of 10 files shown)


designer/server/src/routes/forms/privacy-notice.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+      }
+    }
+  }),
+
+  /**
+   * @satisfies {ServerRoute<{ Params: { slug: string }, Payload: { privacyNoticeUrl : string } }>}
+   */
+  ({
+    method: 'post',
```
alexluckett, line 61: Uppercase for consistency?

alexluckett (approved):

designer/server/src/routes/forms/privacy-notice.js
davidjamesstone: Thanks, see aaa03d85c311230081381531d9050d39fddee847.

There are other places where we use lowercase method in `forms-designer`, but they're `.test` files so I left them.

PR #457, opened 2024-09-09
Add types for form contact details

davidjamesstone:


alexluckett (approved):

model/src/form/form-metadata/index.ts
```diff
@@ -41,6 +41,33 @@ export const privacyNoticeUrlSchema = Joi.string()
...
+  })
+  .trim()
+  .required()
+export const onlineLinkTextSchema = Joi.string().trim().required()
+export const onlineSchema = Joi.object().keys({
+  url: onlineLinkUrlSchema,
+  text: onlineLinkTextSchema
+})
+
+export const contactSchema = Joi.object().keys({
```
colinrotherham, line 65: Should we pass the expected types into Joi?

It's not the most advanced feature, but it'll shout at us if we miss new keys in future etc

```suggestion
export const contactSchema = Joi.object<FormMetadataContact>().keys({
```

davidjamesstone: Resolved 68b3c9ec2f7f8c6b4775cf740ee46f4716d14792

colinrotherham: Happy to leave the other objects for now?

davidjamesstone: For the child schema? c22fd5e999a68152ec20ecd73350eca0245a0232

colinrotherham: Cheers, think you've got them all now

colinrotherham (approved):

PR #458, opened 2024-09-09
Add test coverage for component list and additional settings

colinrotherham:
This PR adds test coverage for:

1. Create component list (by page type)
2. Create or edit component "Additional settings"

Which found a few bugs and issues:

* Fix typo on number field label `for` attribute
* Fix number of rows to use `type="number"`
* Fix punctuation in i18n text (e.g. removed full stops in hints, curly quotes)


designer/client/src/ComponentTypeEdit.test.tsx
```diff
@@ -393,7 +393,7 @@ describe('ComponentTypeEdit', () => {
         const $input = screen.queryByRole<HTMLInputElement>('textbox', {
           name: 'Component name',
           description:
-            'This is generated automatically and does not show on the page. Only change it if you are using an integration that requires you to, for example GOV.UK Notify. It must not contain spaces.'
+            'This is generated automatically and does not show on the page. Only change it if you are using an integration that requires you to, for example GOV.UK Notify. It must not contain spaces'
```
alexluckett, line 396: Are you sure about these? I thought full sentences should retain trailing punctuation and only small one-liners strip them.

alexluckett (approved):

designer/client/src/ComponentTypeEdit.test.tsx
colinrotherham: Ah have you found some updated guidance?

All the Design System examples lack full stops and any [rogue ones were removed](https://github.com/alphagov/govuk-design-system/pull/2077)

There were some "hint" named entries but these were actually just paragraphs in light grey. I've made sure these ones retain trailing punctuation and only actual form component hint text has them dropped

>**When not to use hint text**
>Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.

alexluckett: Those examples are single sentences (they call them "fragment sentences" in the PR you linked), whereas the trailing full stops you've removed are from are paragraphs and not fragment sentences. I get that hint text isn't supposed to contain lengthy explanations, but nonetheless they do currently.

I'm no authority on grammar though which is why I'm asking if you're sure. Better to confirm with the content folks - if they're happy then I'm all good. The PR is approved so you can merge it in once confirmed.

alexluckett: To be clear - it's these multi-sentence hints that I'm questioning the removal for, not single or fragment sentences.

colinrotherham: Glad we spotted these, all sorted now

PR #483, opened 2024-09-20
Add environment identification tag in the header (409424)

davidjamesstone:
Adds an environment tag to the header

<br>

Header tag development

<br>
<br>

Header tag performance test

<br>
<br>

Header tag production

### DIFF
--- a/designer/server/src/common/components/service-header/template.njk
+++ b/designer/server/src/common/components/service-header/template.njk
@@ -1,5 +1,3 @@
-{% from "govuk/components/tag/macro.njk" import govukTag %}
-
 {#
 Component options:
   - organisationName (string): The name of the organisation
@@ -74,8 +72,6 @@ Component options:
             {% endif %}
           </span>
         </a>
-        {# Environment identification tag #}
-        {{ govukTag(params.envTag) }}
       </div>
       <button type="button"
         aria-controls="one-login-header__nav"
--- a/designer/server/src/common/components/service-header/template.njk
+++ b/designer/server/src/common/components/service-header/template.njk
@@ -1,3 +1,5 @@
+{% from "govuk/components/tag/macro.njk" import govukTag %}
+
 {#
 Component options:
   - organisationName (string): The name of the organisation
@@ -72,6 +74,8 @@ Component options:
             {% endif %}
           </span>
         </a>
+        {# Environment identification tag #}
+        {{ govukTag(params.envTag) }}
       </div>
       <button type="button"
         aria-controls="one-login-header__nav"
--- a/designer/server/src/common/components/tag-env/template.njk
+++ b/designer/server/src/common/components/tag-env/template.njk
@@ -1,0 +1,30 @@
+{%- from "govuk/components/tag/macro.njk" import govukTag -%}
+
+{%- switch params.env %}
+  {% case "local" %}
+    {% set text = "Local" %}
+    {% set classes = "govuk-tag--green" %}
+  {% case "dev" %}
+    {% set text = "Development" %}
+    {% set classes = "govuk-tag--grey" %}
+  {% case "test" %}
+    {% set text = "Test" %}
+    {% set classes = "govuk-tag--yellow" %}
+  {% case "ext-test" %}
+    {% set text = "External test" %}
+    {% set classes = "govuk-tag--yellow" %}
+  {% case "perf-test" %}
+    {% set text = "Performance test" %}
+    {% set classes = "govuk-tag--yellow" %}
+  {% case "prod" %}
+    {% set text = "Production" %}
+    {% set classes = "govuk-tag--red" %}
+  {% default %}
+    {% set text = env | replace("-", " ") | capitalize %}
+    {% set classes = "govuk-tag--grey" %}
+{% endswitch -%}
+
+{{ govukTag({
+  text: text,
+  classes: classes + " app-tag--env"
+}) }}
--- a/designer/server/src/common/nunjucks/context/index.js
+++ b/designer/server/src/common/nunjucks/context/index.js
@@ -8,7 +8,7 @@ import { buildNavigation } from '~/src/common/nunjucks/context/build-navigation.
 import config from '~/src/config.js'
 
 const logger = createLogger()
-const { cdpEnvironment, phase, serviceName } = config
+const { cdpEnvironment, phase, serviceName, serviceVersion } = config
 
 /** @type {Record<string, string> | undefined} */
 let webpackManifest
@@ -35,7 +35,8 @@ export async function context(request) {
     config: {
       cdpEnvironment,
       phase,
-      serviceName
+      serviceName,
+      serviceVersion
     },
     navigation: buildNavigation(request),
     getAssetPath: (asset = '') => `/${webpackManifest?.[asset] ?? asset}`,
--- a/designer/server/src/common/nunjucks/context/index.js
+++ b/designer/server/src/common/nunjucks/context/index.js
@@ -1,8 +1,6 @@
 import { readFileSync } from 'node:fs'
 import { basename, join } from 'node:path'
 
-import upperFirst from 'lodash/upperFirst.js'
-
 import { SCOPE_READ } from '~/src/common/constants/scopes.js'
 import { getUserSession } from '~/src/common/helpers/auth/get-user-session.js'
 import { createLogger } from '~/src/common/helpers/logging/logger.js'
@@ -10,7 +8,7 @@ import { buildNavigation } from '~/src/common/nunjucks/context/build-navigation.
 import config from '~/src/config.js'
 
 const logger = createLogger()
-const { phase, serviceName, env, isTest, isProduction } = config
+const { cdpEnvironment, phase, serviceName } = config
 
 /** @type {Record<string, string> | undefined} */
 let webpackManifest
@@ -32,25 +30,10 @@ export async function context(request) {
 
   const credentials = request ? await getUserSession(request) : undefined
 
-  let tagColour
-
-  if (isTest) {
-    tagColour = 'yellow'
-  } else if (isProduction) {
-    tagColour = 'red'
-  } else {
-    tagColour = 'grey'
-  }
-
-  const envTag = {
-    text: upperFirst(env),
-    classes: `govuk-tag--${tagColour}`
-  }
-
   return {
-    envTag,
     breadcrumbs: [],
     config: {
+      cdpEnvironment,
       phase,
       serviceName
     },
--- a/designer/server/src/common/nunjucks/context/index.js
+++ b/designer/server/src/common/nunjucks/context/index.js
@@ -1,6 +1,8 @@
 import { readFileSync } from 'node:fs'
 import { basename, join } from 'node:path'
 
+import upperFirst from 'lodash/upperFirst.js'
+
 import { SCOPE_READ } from '~/src/common/constants/scopes.js'
 import { getUserSession } from '~/src/common/helpers/auth/get-user-session.js'
 import { createLogger } from '~/src/common/helpers/logging/logger.js'
@@ -8,7 +10,7 @@ import { buildNavigation } from '~/src/common/nunjucks/context/build-navigation.
 import config from '~/src/config.js'
 
 const logger = createLogger()
-const { phase, serviceName } = config
+const { phase, serviceName, env, isTest, isProduction } = config
 
 /** @type {Record<string, string> | undefined} */
 let webpackManifest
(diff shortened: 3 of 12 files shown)


designer/server/src/common/components/service-header/template.njk
```diff
@@ -72,6 +74,8 @@ Component options:
             {% endif %}
           </span>
         </a>
+        {# Environment identification tag #}
```
colinrotherham, line 77: Super minor but this comment will inject a new line

Many many more haven't been tackled so not a blocker in the slightest

```suggestion
        {# Environment identification tag -#}
```

designer/server/src/common/templates/partials/navigation.njk
```diff
@@ -4,6 +4,7 @@
 {{ appServiceHeader({
   organisationName: "Defra",
   productName: "Forms Designer",
+  envTag: envTag,
```
colinrotherham, line 7: Shall we use Nunjucks for what it's good at?

By extending the [`config` context](https://github.com/DEFRA/forms-designer/blob/main/designer/server/src/common/nunjucks/context/index.js#L35-L38) we can do `config.env | titlecase` from inside a macro?

```njk
{%- macro envTag(env) -%}
  {%- switch env %}
    {% case "test" %}
      {% set classes = "govuk-tag--yellow" %}
    {% case "production" %}
      {% set classes = "govuk-tag--red" %}
    {% default %}
      {% set classes = "govuk-tag--grey" %}
  {% endswitch -%}

  {{ govukTag(
    text: upperFirst(env),
    classes: classes
  }) }}
{%- endmacro -%}
```

Could pass it in and add the macro alongside `littlePersonIcon()` already in the header:

```suggestion
  env: config.env,
```

davidjamesstone: I'm easy but honestly I'm not convinced with the njk approach.

`njk title filter != upper case first letter` (looks like it [upper cases every word](https://github.com/mozilla/nunjucks/blob/master/nunjucks/src/filters.js#L508) like we saw a few months back)

Then there's the use of the undocumented jinja `switch` case statement (I wish they'd get round to this).

colinrotherham: Yeah assumed our environments were only single words

We've used `switch()` here already if it helps?

https://github.com/DEFRA/forms-designer/blob/f56d3af386f21db3056220ff0a2574fe8c35d4be/designer/server/src/common/components/heading/template.njk#L5-L15

davidjamesstone: It makes me uncomfortable that it's not documented.

Also, I'm in the logic-less templates school of thought ala Handlebars so IMO, less NJK/template logic is better than more. JS will be around long after nunjucks has [been replaced](https://github.com/mozilla/nunjucks/issues/1433) or GDS decide to use something else.

Honestly though, I'm happy to go either way.

colinrotherham: I'm happy to swap it for if/else but for now the Nunjucks ship has sailed

For the next developer that joins, let's keep header logic inside the header

designer/server/src/common/nunjucks/context/index.js
```diff
@@ -29,8 +30,10 @@ export async function context(request) {
   }
 
   const credentials = request ? await getUserSession(request) : undefined
+  const envTag = environmentTagHelper(env, isTest, isProduction)
```
alexluckett, line 33: Seems a bit excessive to pass in both the env tag and the flags? Wouldn't `env` cover this on its own?

designer/server/src/common/nunjucks/context/index.js
```diff
@@ -3,12 +3,13 @@ import { basename, join } from 'node:path'
...
 import { SCOPE_READ } from '~/src/common/constants/scopes.js'
 import { getUserSession } from '~/src/common/helpers/auth/get-user-session.js'
+import environmentTagHelper from '~/src/common/helpers/environment-tag-helper.js'
 import { createLogger } from '~/src/common/helpers/logging/logger.js'
 import { buildNavigation } from '~/src/common/nunjucks/context/build-navigation.js'
 import config from '~/src/config.js'
 
 const logger = createLogger()
-const { phase, serviceName } = config
+const { phase, serviceName, env, isTest, isProduction } = config
```
alexluckett, line 12: `config.env` is only `development`/`test` locally. In a live environment it always runs as `production`.

designer/server/src/common/templates/partials/navigation.njk
```diff
@@ -4,6 +4,7 @@
 {{ appServiceHeader({
   organisationName: "Defra",
   productName: "Forms Designer",
+  env: env,
```
colinrotherham, line 7: Might as well pass `config` here so it's easier to search for `{{ config.env }}`

Or if you're happy that `config` is already in the global context, you don't even need a param

```suggestion
  config: config,
```

designer/server/src/common/nunjucks/context/index.js
```diff
@@ -31,6 +31,7 @@ export async function context(request) {
   const credentials = request ? await getUserSession(request) : undefined
 
   return {
+    env,
     breadcrumbs: [],
     config: {
       phase,
```
colinrotherham, line 37: Mind sticking this in `config` so it's still `{{ config.env }}` in Nunjucks too?

```suggestion
    breadcrumbs: [],
    config: {
      env,
      phase,
```

colinrotherham (commented):
Thanks @KiranVarma02

Are you happy everything works without needing tests?

You can copy the Nunjucks tests from another component if you like:
https://github.com/DEFRA/forms-designer/blob/main/designer/server/src/common/components/heading/template.test.js

designer/server/src/common/nunjucks/context/index.js
KiranVarma02: agree

designer/server/src/common/templates/partials/navigation.njk
KiranVarma02: config is not in global context. I am passing just env for now. If we need more from config it can be changed in future.

designer/server/src/common/components/service-header/template.njk
```diff
@@ -1,3 +1,5 @@
+{% from "service-header/environment-tag.njk" import envTag %}
```
colinrotherham, line 1: Looking good @KiranVarma02

Probably want to split this out from the service header if it's going in the GOV.UK header too?

I'd prefix it with "app" as `appTagEnv()` like the others, and use the `macro.njk` + `template.njk` filenames

Then I think it's job done

```suggestion
{% from "tag-env/macro.njk" import appTagEnv %}
```

KiranVarma02: done, I like it. Thanks @colinrotherham

designer/server/src/common/components/tag-env/template.njk
```diff
@@ -0,0 +1,15 @@
...
+    {% case "production" %}
+      {% set classes = "govuk-tag--red" %}
+    {% default %}
+      {% set classes = "govuk-tag--grey" %}
+  {% endswitch -%}
+
+  {{ govukTag({
+    text: env | capitalize,
+    classes: classes
+  }) }}
```
colinrotherham, line 15: Mind if we fix the indentation here?

```suggestion
  {% case "test" %}
    {% set classes = "govuk-tag--yellow" %}
  {% case "production" %}
    {% set classes = "govuk-tag--red" %}
  {% default %}
    {% set classes = "govuk-tag--grey" %}
{% endswitch -%}

{{ govukTag({
  text: env | capitalize,
  classes: classes
}) }}
```

designer/server/src/common/components/tag-env/template.njk
```diff
@@ -0,0 +1,15 @@
...
+    {% case "test" %}
+      {% set classes = "govuk-tag--yellow" %}
+    {% case "production" %}
+      {% set classes = "govuk-tag--red" %}
+    {% default %}
+      {% set classes = "govuk-tag--grey" %}
+  {% endswitch -%}
+
+  {{ govukTag({
+    text: env | capitalize,
```
colinrotherham, line 13: If we use the `replace` filter we can keep `perf-test` (hyphenated) as the env name

```suggestion
    text: env | replace("-", " ") | capitalize,
```

designer/server/src/config.ts
```diff
@@ -37,14 +38,18 @@ export interface Config {
 const schema = joi.object<Config>({
   port: joi.number().default(3000),
   env: joi
+    .string()
+    .valid('development', 'test', 'perf test', 'production')
```
colinrotherham, line 42: I'm still thinking we should match these names to how they're used in URLs

We might want to use them to replace wildcards in URLs in future

```suggestion
    .valid('dev', 'test', 'perf-test', 'prod')
```

designer/server/src/config.ts
```diff
@@ -88,15 +93,18 @@ const schema = joi.object<Config>({
...
     previewUrl: process.env.PREVIEW_URL,
     serviceName: 'Submit a form to Defra',
     logLevel: process.env.LOG_LEVEL,
     phase: process.env.PHASE,
-    isProduction: process.env.NODE_ENV === 'production',
-    isDevelopment: !['production', 'test'].includes(`${process.env.NODE_ENV}`),
+    isProduction: process.env.DEPLOYMENT_ENV === 'production',
+    isDevelopment: !['production', 'test'].includes(
+      `${process.env.DEPLOYMENT_ENV}`
+    ),
```
colinrotherham, line 107: When `DEPLOYMENT_ENV` is missing do we definitely want to run `.includes(undefined)`?

Might need to do this?

```suggestion
    isDevelopment: !process.env.DEPLOYMENT_ENV || !['production', 'test'].includes(
      process.env.DEPLOYMENT_ENV
    ),
```

<br>

If you want to keep code size down, can do an all-in-one destructure at the top?

```js
const {
  DEPLOYMENT_ENV,
  LOG_LEVEL,
  NODE_ENV,
  PHASE,
  SESSION_COOKIE_PASSWORD,
  SESSION_TTL,
  // … more
} = process.env
```

designer/server/src/config.ts
KiranVarma02: I will let @alexluckett decide that I think! 

This is for non production envs, so I don't have strong opinion

colinrotherham: I'll ask on Slack

What shall we use when running locally?

Because the "dev" environment isn't local, e.g.

http://forms-designer.dev.cdp-int.defra.cloud
http://forms-runner.dev.cdp-int.defra.cloud

We'll need maybe `local` or `localhost` as the default

designer/server/src/common/components/tag-env/template.njk
colinrotherham: Yeah I've asked on Slack, doing that prevents use in URLs:

```js
new URL(`http://forms-runner/.${config.env}.cdp-int.defra.cloud`)
```

designer/server/src/common/components/tag-env/template.njk
```diff
@@ -0,0 +1,15 @@
+{%- from "govuk/components/tag/macro.njk" import govukTag -%}
+
+{%- switch env %}
+    {% case "test" %}
+      {% set classes = "govuk-tag--yellow" %}
+    {% case "production" %}
+      {% set classes = "govuk-tag--red" %}
+    {% default %}
+      {% set classes = "govuk-tag--grey" %}
```
colinrotherham, line 9: Sorry @KiranVarma02 these inner lines still need indenting

```suggestion
  {% case "test" %}
    {% set classes = "govuk-tag--yellow" %}
  {% case "production" %}
    {% set classes = "govuk-tag--red" %}
  {% default %}
    {% set classes = "govuk-tag--grey" %}
```

KiranVarma02: Done. If I am working on this project for longer, I would add lint rule for this. 

Maybe I will look for it this afternoon.

designer/server/src/config.ts
```diff
@@ -89,31 +118,32 @@ const schema = joi.object<Config>({
...
-    redisHost: process.env.REDIS_HOST,
-    redisUsername: process.env.REDIS_USERNAME,
-    redisPassword: process.env.REDIS_PASSWORD,
-    redisKeyPrefix: process.env.REDIS_KEY_PREFIX,
-    roleEditorGroupId: process.env.ROLE_EDITOR_GROUP_ID
+    logLevel: LOG_LEVEL,
+    phase: PHASE,
+    isProduction: DEPLOYMENT_ENV === 'prod',
+    isDevelopment:
+      !DEPLOYMENT_ENV || !['prod', 'test'].includes(DEPLOYMENT_ENV),
```
KiranVarma02, line 132: Is this still what we want?

colinrotherham: Yeah think so

CDP do have the description "**If this application running in the development environment**" but they similarly set it to `process.env.NODE_ENV !== 'production'` so I'm happy with your changes

i.e. You've kept the same behaviour yet we can adjust it in future if needed

alexluckett (approved):

alexluckett (approved):

PR #494, opened 2024-09-23
Add list type radio selection to the form

davidjamesstone:
Screenshot from 2024-11-07 16-00-38

Closing this for now while there isn't a need for it.
The code in defra/forms-model and forms-runner to support numeric lists will remain but we'll pick up the designer part if a need for numeric lists ever arises, at which point we expect it'll be part of the new (non-reactjs) editor

### DIFF
--- a/designer/client/src/hooks/list/useListItem/useListItem.tsx
+++ b/designer/client/src/hooks/list/useListItem/useListItem.tsx
@@ -1,4 +1,4 @@
-import { type Root } from 'joi'
+import Joi, { type Root } from 'joi'
 
 import { findListItem } from '~/src/data/list/findList.js'
 import { type ListItemHook } from '~/src/hooks/list/useListItem/types.js'
--- a/designer/client/src/hooks/list/useListItem/useListItem.tsx
+++ b/designer/client/src/hooks/list/useListItem/useListItem.tsx
@@ -1,4 +1,4 @@
-import Joi, { type Root } from 'joi'
+import { type Root } from 'joi'
 
 import { findListItem } from '~/src/data/list/findList.js'
 import { type ListItemHook } from '~/src/hooks/list/useListItem/types.js'
@@ -93,7 +93,7 @@ export function useListItem(
       errors.value ??= validateCustom('value', value, {
         message: 'errors.number',
         label: `Item value '${value}'`,
-        schema: Joi.number().required()
+        schema: schema.number().required()
       })
     }
 
--- a/designer/client/src/hooks/list/useListItem/useListItem.tsx
+++ b/designer/client/src/hooks/list/useListItem/useListItem.tsx
@@ -89,6 +89,14 @@ export function useListItem(
       schema
     })
 
+    if (selectedList?.type === 'number') {
+      errors.value ??= validateCustom('value', value, {
+        message: 'errors.number',
+        label: `Item value '${value}'`,
+        schema: Joi.number().required()
+      })
+    }
+
     errors.value ??= validateCustom('value', [...values, value], {
       message: 'errors.duplicate',
       label: `Item value '${value}'`,
--- a/designer/client/src/list/ListEdit.test.tsx
+++ b/designer/client/src/list/ListEdit.test.tsx
@@ -40,7 +40,7 @@ describe('ListEdit', () => {
 
     const $listCaption = screen.getByText('List items')
     const $listHint = screen.getByText('Enter a unique name for your list')
-    const $listType = screen.getByText('Value type for your list')
+    const $listType = screen.getByText('List type')
     const $listLink = screen.getByText('Add a new list item')
 
     expect($listCaption).toBeInTheDocument()
--- a/designer/client/src/list/ListEdit.test.tsx
+++ b/designer/client/src/list/ListEdit.test.tsx
@@ -40,10 +40,12 @@ describe('ListEdit', () => {
 
     const $listCaption = screen.getByText('List items')
     const $listHint = screen.getByText('Enter a unique name for your list')
+    const $listType = screen.getByText('Value type for your list')
     const $listLink = screen.getByText('Add a new list item')
 
     expect($listCaption).toBeInTheDocument()
     expect($listHint).toBeInTheDocument()
+    expect($listType).toBeInTheDocument()
     expect($listLink).toBeInTheDocument()
   })
 })
--- a/designer/client/src/list/ListEdit.tsx
+++ b/designer/client/src/list/ListEdit.tsx
@@ -246,6 +246,9 @@ export function ListEdit() {
               <legend className="govuk-fieldset__legend govuk-fieldset__legend--s">
                 {i18n('list.type')}
               </legend>
+              <div id="list-type-hint" className="govuk-hint">
+                {i18n('list.typeHint')}
+              </div>
               <div
                 id="list-type"
                 className="govuk-radios govuk-radios--small"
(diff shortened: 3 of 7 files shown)


designer/client/src/list/ListEdit.tsx
```diff
@@ -193,6 +193,44 @@ export function ListEdit() {
...
+          value={selectedList.type}
+          defaultValue="string"
+          items={[
+            {
+              children: ['String'],
+              value: 'string',
+              hint: {
+                children: [i18n('list.typeHintString')]
+              },
+              disabled: !selectedList.isNew
```
colinrotherham, line 215: Thanks for pushing this up Dave

Since screen readers can't focus + announce disabled form fields, can we change this to text?

E.g. Just a heading and a description instead:

> **List type**
> Text values only

> **List type**
> Numeric values only

Save the non-disabled form fields for when creating new lists only

Let me know if you need a hand running it through `validate()`

designer/client/src/list/ListEdit.tsx
```diff
@@ -193,6 +193,44 @@ export function ListEdit() {
...
+          name="type"
+          hint={{ children: i18n('list.typeHint') }}
+          fieldset={{
+            legend: {
+              className: 'govuk-fieldset__legend--s',
+              children: [i18n('list.type')]
+            }
+          }}
+          value={selectedList.type}
+          defaultValue="string"
```
colinrotherham, line 207: We always have `selectedList.type` (and it's updated on change) so we'd best remove `defaultValue`

```suggestion
```

Probably seen this in the browser console?

>Warning: Boolean contains an input of type radio with both checked and defaultChecked props. Input elements must be either controlled or uncontrolled (specify either the checked prop, or the defaultChecked prop, but not both)

designer/client/src/reducers/listReducer.tsx
```diff
@@ -44,6 +49,7 @@ export type ListReducerActions =
   | {
       name:
         | ListActions.EDIT_TITLE
+        | ListActions.EDIT_TYPE
```
colinrotherham, line 52: Thanks for adding this new action

Could you split it out but typed as `ListTypeContent` instead of string?

To make sure we surface errors for non-strings in **ListEdit.tsx**

designer/client/src/reducers/listReducer.tsx
```diff
@@ -132,6 +138,10 @@ export function listReducer(state: ListState, action: ListReducerActions) {
       selectedList.title = payload
       break
 
+    case ListActions.EDIT_TYPE:
+      selectedList.type = payload as ListTypeContent
```
colinrotherham, line 142: Would you mind if we moved the `as ListTypeContent` to the `onChange` handler?

We do something similar when changing page controller:

https://github.com/DEFRA/forms-designer/blob/98fe937bc3d076823c1707273f3830da2167d4ad/designer/client/src/PageCreate.tsx#L201-L207

designer/client/src/list/ListEdit.tsx
```diff
@@ -193,6 +193,44 @@ export function ListEdit() {
           }
           errorMessage={errors.title}
         />
+        <Radios
```
colinrotherham, line 196: Not your fault but the Xgov `<Radios>` component isn't compatible with GOV.UK Frontend v5x

```console
Unhandled Promise Rejection: TypeError: undefined is not a constructor (evaluating 'new RadiosJS(controlRef.current)')
```

Probably one for the backlog, but until we migrate away we can use the radio HTML directly like in:

https://github.com/DEFRA/forms-designer/blob/98fe937bc3d076823c1707273f3830da2167d4ad/designer/client/src/components/FieldEditors/ListContentEdit.tsx#L29-L35

designer/client/src/hooks/list/useListItem/useListItem.tsx
```diff
@@ -89,6 +89,14 @@ export function useListItem(
       schema
     })
 
+    if (selectedList?.type === 'number') {
+      errors.value ??= validateCustom('value', value, {
+        message: 'errors.number',
+        label: `Item value '${value}'`,
+        schema: Joi.number().required()
```
colinrotherham, line 96: This is looking ace @davidjamesstone

I'm using dependency injection for Joi now so if you can remove `import Joi` and change this line?

Then webpack can lazy-load `/javascripts/vendor/joi.[hash].min.js` via https://github.com/DEFRA/forms-designer/commit/1c3370aec473c00402c680c116be7c6fc599f55a

Hadn't realised the Joi browser build was 144KB (see https://github.com/DEFRA/forms-designer/pull/497 for info)

```suggestion
        schema: schema.number().required()
```

davidjamesstone: Not sure how I missed that. During the rebase maybe.

Fixed in 2331da9c3618413dada7df4aaa1c215488cb7f21

davidjamesstone:
I wasn't too thrilled with only allowing the type to change if listItems.length is zero but it ducks a bunch of potential number <=> string coercion issues.

designer/client/src/hooks/list/useListItem/useListItem.tsx
davidjamesstone: Yeah joi.browser is a bit big, I believe they already strip a bunch of email/URL logic out to bring the file size down a bit.

designer/client/src/list/ListEdit.tsx
```diff
@@ -200,6 +213,73 @@ export function ListEdit() {
...
+              {`${
+                listTypeRadioItems.find(
+                  (item) => item.value === selectedList.type
+                )?.labelText
+              } values only`}
+            </p>
+          </div>
+        ) : (
+          <div className="govuk-form-group">
+            <fieldset className="govuk-fieldset">
```
colinrotherham, line 231: Mind just wiring up the hint ID for me please?

Hadn't realised **ListEdit.test.tsx** wasn't switched to role tests so this got missed

```suggestion
            <fieldset
              className="govuk-fieldset"
              aria-describedby="list-type-hint"
            >
```

designer/client/src/list/ListEdit.test.tsx
```diff
@@ -40,10 +40,12 @@ describe('ListEdit', () => {
 
     const $listCaption = screen.getByText('List items')
     const $listHint = screen.getByText('Enter a unique name for your list')
+    const $listType = screen.getByText('List type')
```
colinrotherham, line 43: Not a blocker, but can we add role tests for the radios at some point?

E.g. Here's what we do in **ListContentEdit.test.tsx**

https://github.com/DEFRA/forms-designer/blob/2331da9c3618413dada7df4aaa1c215488cb7f21/designer/client/src/components/FieldEditors/ListContentEdit.test.tsx#L87-L104

designer/client/src/list/ListEdit.tsx
```diff
@@ -200,6 +213,73 @@ export function ListEdit() {
...
+                id="list-type"
+                className="govuk-radios govuk-radios--small"
+                data-module="govuk-radios"
+              >
+                {listTypeRadioItems.map((item) => {
+                  const name = 'type'
+                  const id = `${name}-${item.value}`
+
+                  return (
+                    <div className="govuk-radios__item" key={item.value}>
```
colinrotherham, line 248: React can cache and skip re-rendering by key

Is using the value definitely unique enough?

```suggestion
                    <div className="govuk-radios__item" key={`${list.name}-${id}`}>
```

PR #500, opened 2024-09-24
Feature/342247 email address validation

davidjamesstone:


KiranVarma02 (approved):

designer/server/src/routes/forms/notification-email.js
```diff
@@ -14,9 +14,12 @@ export const ROUTE_PATH_EDIT_NOTIFICATION_EMAIL =
...
 export const schema = Joi.object().keys({
   notificationEmail: emailAddressSchema.required().messages({
-    'string.empty': 'Enter an email address submitted forms should be sent to',
+    'string.empty':
+      'Enter an email address that submitted forms should be sent to',
     'string.email':
-      'Enter an email address submitted forms should be sent to in the correct format, like name@gov.uk'
+      'Enter an email address that submitted forms should be sent to in the correct format',
+    'string.pattern.base':
+      'Enter an email address that submitted forms should be sent to in the correct format, like name@example.gov.uk or name@example.org.uk'
```
colinrotherham, line 22: Just moving the feedback from Slack so it doesn't get lost

We've got two ways examples are explained

### Lede text
>…must end with ‘.gov.uk’ or ‘.org.uk’.

### Error message
>…like name@example.gov.uk or name@example.org.uk

We'll want to confirm if we should

* make the examples consistent
* consolidate the examples into hint text

```suggestion
      'Enter an email address that submitted forms should be sent to in the correct format'
```

colinrotherham: For example the guidance for error messages under [**Be clear and concise**](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise) says:

>Do not give an example in the error message if there is an example on the screen. For example, if you are asking for a National Insurance number and include ‘QQ 12 34 56 C’ as hint text, do not include an example in the error message.

For visual and assistive software context, we should probably move our lede examples into hint text

>**If the email address is not in the correct format and there is no example**
>
>Say ‘Enter an email address in the correct format, like name@example.com’.
>
>**If the email address is not in the correct format and there is an example**
>
>Say ‘Enter an email address in the correct format’.

PR #509, opened 2024-09-26
feat: [435701] Remember password file download

KiranVarma02:


designer/server/src/routes/file/file.js
```diff
@@ -66,13 +76,23 @@ export default [
...
 
       try {
         const { url } = await createFileLink(fileId, email, token)
+
+        await server.methods.session.set(
+          sessionNames.fileDownloadPassword,
+          email,
+          {
+            cache: {
+              expiresIn: config.fileDownloadPasswordTtl
```
alexluckett, line 92: 👌

designer/server/src/types.ts
```diff
@@ -28,7 +32,7 @@ declare module '@hapi/hapi' {
   interface Server {
     method(name: 'session.drop', method: SessionCache['drop']): void
     method(name: 'session.get', method: SessionCache['get']): void
-    method(name: 'session.set', method: SessionCache['set']): void
+    method(name: 'session.set', method: SessionCache['set'], ttl?: number): void
```
colinrotherham, line 35: Looks like we're not using the 3rd param, can you double check?

It's the `method` function that gets the TTL as the 3rd param

designer/server/src/config.ts
```diff
@@ -71,6 +72,7 @@ const schema = joi.object<Config>({
   isDevelopment: joi.boolean().default(true),
   isTest: joi.boolean().default(false),
   sessionTtl: joi.number().required(),
+  fileDownloadPasswordTtl: joi.number().default(34560),
```
alexluckett, line 75: Can you remove the default value and make it required please? Raise a PR for https://github.com/DEFRA/cdp-app-config/ with the values for each environment.

In that repo you can set it per environment or a global value. Use 2 days for dev, then the keep the other environments as per the ticket.

designer/server/src/types.ts
KiranVarma02: you are right. Removed it.

designer/server/src/createServer.ts
```diff
@@ -84,6 +84,16 @@ export async function createServer() {
...
 
+  server.method('state.get', (userId, key) => {
+    if (!userId) return Promise.reject(new Error('userId is not defined'))
+    return cache.get(`${userId}.${key}`)
+  })
+  server.method('state.set', (userId, key, value, ttl = config.sessionTtl) => {
+    if (!userId) return Promise.reject(new Error('userId is not defined'))
+    return cache.set(`${userId}.${key}`, value, ttl)
+  })
+  server.method('state.drop', (id) => cache.drop(id))
```
colinrotherham, line 95: I know we're not using it and it's only here for completeness, but the key needs updating?

```suggestion
  server.method('state.drop', (userId, key) => cache.drop(`${userId}.${key}`))
```

designer/server/src/createServer.ts
```diff
@@ -84,6 +84,16 @@ export async function createServer() {
   server.method('session.set', (id, value) => cache.set(id, value))
   server.method('session.drop', (id) => cache.drop(id))
 
+  server.method('state.get', (userId, key) => {
+    if (!userId) return Promise.reject(new Error('userId is not defined'))
```
colinrotherham, line 88: Can I check when the `userId` might be empty?

If you've found a way this is possible we should update `hasUser()` instead

Otherwise can probably remove the checks

colinrotherham (commented):
Thanks @KiranVarma02

Added a few comments whilst we decide on what key to use

Do you think we should make the `state.*` methods generic so you pass in a prefix rather than only a user ID?

```js
server.method('session.get', (...args) => cache.get(...args))
server.method('session.set', (...args) => cache.set(...args))
server.method('session.drop', (...args) => cache.drop(...args))

server.method('state.get', (key, options) =>
  cache.get(`${options.prefix}:${key}`)
)

server.method('state.set', (key, value, options) =>
  cache.set(`${options.prefix}:${key}`, value, options.ttl)
)

server.method('state.drop', (key, value, options) =>
  cache.drop(`${options.prefix}:${key}`)
)
```

designer/server/src/createServer.ts
KiranVarma02: I am not sure in real scenario but for testing I have removed auth and userId doesn't exist in that case

colinrotherham: I'm trying to avoid writing code we can't test and is likely never run

E.g. With the [suggestion above](https://github.com/DEFRA/forms-designer/pull/509#discussion_r1783976070) is it definitely possible to have no user ID?

```js
if (!hasUser(credentials)) {
  return Boom.unauthorized()
}
```

designer/server/src/routes/file/file.js
```diff
@@ -33,7 +35,19 @@ export default [
...
+
+          const email =
+            (await server.methods.state.get(
+              userId,
+              sessionNames.fileDownloadPassword
+            )) ?? ''
+          return h.view(
+            'file/download-page',
+            file.fileViewModel(email, validation)
+          )
```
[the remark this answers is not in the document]
KiranVarma02: I like it in a way but I am not sure if its "clean" to add normal data to validation object. I am going to make the change anyway, let me know what you think!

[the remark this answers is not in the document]
colinrotherham: That's fair. Alternatively an entire object would be consistent with other view models

E.g. Like passing `metadata` into `titleViewModel(metadata, validation)` we could pass in `file`

```patch
-  file.fileViewModel(validation)
+  file.fileViewModel(file, session)
```

Giving us access to `file.password` and `file.XXX` in future

Either works for me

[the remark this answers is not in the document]
KiranVarma02: I think I will leave it as it is now but I agree adding file if we need more data is good

designer/server/src/createServer.ts
KiranVarma02: after adding above Boom return we don't need error in createserver.js, I will remove

designer/server/src/routes/file/file.js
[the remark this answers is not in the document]
KiranVarma02: I am having other issues with below when validation is null. I am reverting to what it was which seems better instead of making more changes

`if (email && !validation.formValues.email) {
            validation.formValues.email = email
          }`

KiranVarma02:
> Thanks @KiranVarma02
> 
> Added a few comments whilst we decide on what key to use
> 
> Do you think we should make the `state.*` methods generic so you pass in a prefix rather than only a user ID?
> 
> ```js
> server.method('session.get', (...args) => cache.get(...args))
> server.method('session.set', (...args) => cache.set(...args))
> server.method('session.drop', (...args) => cache.drop(...args))
> 
> server.method('state.get', (key, options) =>
>   cache.get(`${options.prefix}:${key}`)
> )
> 
> server.method('state.set', (key, value, options) =>
>   cache.set(`${options.prefix}:${key}`, value, options.ttl)
> )
> 
> server.method('state.drop', (key, value, options) =>
>   cache.drop(`${options.prefix}:${key}`)
> )
> ```

we could but if there are changes in args we have to update types anyway. So, I am not sure.

designer/server/src/createServer.ts
```diff
@@ -84,6 +84,12 @@ export async function createServer() {
   server.method('session.set', (id, value) => cache.set(id, value))
   server.method('session.drop', (id) => cache.drop(id))
 
+  server.method('state.get', (userId, key) => cache.get(`${userId}.${key}`))
+  server.method('state.set', (userId, key, value, ttl = config.sessionTtl) =>
```
colinrotherham, line 88: Do we definitely need to set the default if it's already the default?

```suggestion
  server.method('state.set', (userId, key, value, ttl) =>
```

colinrotherham:
> we could but if there are changes in args we have to update types anyway. So, I am not sure.

@KiranVarma02 Moving to `options` was to keep everything understandable

We started with "name and value" but by adding two more params it's hard to tell what they are:

### Before
```js
await server.methods.state.set(
  credentials.user.id,
  sessionNames.fileDownloadPassword,
  email,
  config.fileDownloadPasswordTtl
)
```

### After
```js
const options = {
  prefix: credentials.user.id,
  ttl: config.fileDownloadPasswordTtl
}

await server.methods.state.set(
  sessionNames.fileDownloadPassword,
  email,
  options
)
```

designer/server/src/routes/file/file.js
```diff
@@ -41,16 +41,16 @@ export default [
...
-          )
-
-          if (email && !validation.formValues.email) {
-            validation.formValues.email = email
-          }
+          const email =
+            (await server.methods.state.get(
+              credentials.user.id,
+              sessionNames.fileDownloadPassword
+            )) ?? ''
```
colinrotherham, line 48: Did the last refactor revert this?

```suggestion
          const email = await server.methods.state.get(
            credentials.user.id,
            sessionNames.fileDownloadPassword
          )
```

KiranVarma02: refactor done

designer/server/src/createServer.ts
KiranVarma02: I missed this before. true, I agree. I will remove default.

KiranVarma02: done

KiranVarma02:
> > we could but if there are changes in args we have to update types anyway. So, I am not sure.
> 
> @KiranVarma02 Moving to `options` was to keep everything understandable
> 
> We started with "name and value" but by adding two more params it's hard to tell what they are:
> 
> ### Before
> ```js
> await server.methods.state.set(
>   credentials.user.id,
>   sessionNames.fileDownloadPassword,
>   email,
>   config.fileDownloadPasswordTtl
> )
> ```
> 
> ### After
> ```js
> const options = {
>   prefix: credentials.user.id,
>   ttl: config.fileDownloadPasswordTtl
> }
> 
> await server.methods.state.set(
>   sessionNames.fileDownloadPassword,
>   email,
>   options
> )
> ```

I tried but because there are only 2 args, I think not using `...args` seems easy to read for me.

designer/server/src/routes/file/file.test.js
```diff
@@ -52,6 +73,25 @@ describe('File routes', () => {
...
+    test('should show unauthorized page when user is unauthorized', async () => {
+      jest.mocked(file.checkFileStatus).mockResolvedValueOnce(StatusCodes.OK)
+
+      const options = {
+        method: 'GET',
+        url: fileDownloadUrl,
+        auth: { ...auth, credentials: {} }
+      }
+
+      // @ts-expect-error: auth credentials are intentionally empty for this test case
```
KiranVarma02, line 86: this is okay I hope or is there other way to do it?

KiranVarma02:
Tests that I have added definitely tests uncovered lines which sonar says are not covered. I can see it in debug. Is it possible sonar is not running new coverage report? - @colinrotherham

colinrotherham:
@KiranVarma02 I did think it's because Hapi is already preventing requests without auth?

i.e. It never gets chance to return via the `Boom.unauthorized()` line

```js
    options: {
      auth: {
        access: {
          entity: 'user',
          scope: false
        }
      }
    }
```

alexluckett (approved):

designer/server/src/routes/file/file.js
```diff
@@ -23,17 +26,30 @@ export default [
...
 
       switch (statusCode) {
         case StatusCodes.OK: {
           const validation = yar.flash(
             sessionNames.validationFailure.fileDownload
           )[0]
-          return h.view('file/download-page', file.fileViewModel(validation))
+          const email =
+            (await server.methods.state.get(
+              user.id,
```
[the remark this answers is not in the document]
alexluckett: Ignore this, it's from a stale review I abandoned and moved on to Slack.

PR #533, opened 2024-10-03
Remove unnecessary helper `isQuestionPage()`

colinrotherham:
This PR moves **PageEdit.tsx** and **PageCreate.tsx** `getPageDefaults()` into state

By doing this we can remove an unnecessary helper `isQuestionPage()` and use `hasFormComponents()` instead


KiranVarma02 (approved):
LGTM

PR #557, opened 2024-10-18
Add form-submission API types

davidjamesstone:


model/src/form/form-submission/index.ts
```diff
@@ -0,0 +1,42 @@
...
+})
+
+/**
+ * Joi schema for `SubmitRecordset` interface
+ * @see {@link SubmitRecordset}
+ */
+export const formSubmitRecordsetSchema = Joi.object<SubmitRecordset>({
+  name: Joi.string().required(),
+  title: Joi.string().required(),
+  value: Joi.array()
```
colinrotherham, line 26: Do we need to type these arrays like elsewhere?

```suggestion
  value: Joi.array<XXX>()
```

davidjamesstone: Resolved with 9e45b534c7586486c57786c30cc827ac19081d69

colinrotherham (approved):

PR #589, opened 2024-11-08
feat(466411): remove email case sensitivity check

mokhld:
This PR implements case sensitivity handling for file downloads in the Forms Designer service, working in conjunction with the Forms Submission API changes.

1. Notification Email Handling
- All notification emails are now consistently lowercased when:
  - Creating new forms
  - Editing existing forms
- Maintains consistent lowercase storage for new/edited forms
- Handles email case transformation within the Forms Designer service

2. File Downloads
- Added case sensitivity check when validating download emails
- Email validation now considers case sensitivity flag from Forms Submission API
- Implements conditional email transformation based on case sensitivity requirements
- Retrieves `emailIsCaseSensitive` flag from Forms Submission API
- Ensures backward compatibility with existing records

3. Handling Scenarios 🥴
  - New Forms:
    - Notification emails automatically lowercased
    - File downloads use case-insensitive matching

  - Legacy Forms:
    - Respects original email case sensitivity
    - Maintains backward compatibility

  - Mixed Scenarios
    - Handles both case-sensitive and insensitive matching
    - Preserves existing functionality

Sister PR: DEFRA/forms-submission-api#161 for more thorough description of the api changes

### DIFF
--- a/model/src/form/form-metadata/index.ts
+++ b/model/src/form/form-metadata/index.ts
@@ -76,7 +76,6 @@ export const privacyNoticeUrlSchema = Joi.string()
 export const notificationEmailAddressSchema = Joi.string()
   .email({ tlds: { allow: ['uk'] } })
   .trim()
-  .lowercase()
   .pattern(/\.gov\.uk$|\.org\.uk$/)
 
 export const authoredAtSchema = Joi.date().iso().required()
--- a/model/src/form/form-metadata/index.ts
+++ b/model/src/form/form-metadata/index.ts
@@ -43,7 +43,7 @@ export const phoneSchema = Joi.string().trim()
 export const emailAddressSchema = Joi.string().email().trim().required()
 export const emailResponseTimeSchema = Joi.string().trim().required()
 export const emailSchema = Joi.object<FormMetadataContactEmail>().keys({
-  address: emailAddressSchema,
+  address: emailAddressSchema.lowercase(),
   responseTime: emailResponseTimeSchema
 })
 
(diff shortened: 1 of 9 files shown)


model/src/form/form-metadata/index.ts
```diff
@@ -43,7 +43,7 @@ export const phoneSchema = Joi.string().trim()
 export const emailAddressSchema = Joi.string().email().trim().required()
 export const emailResponseTimeSchema = Joi.string().trim().required()
 export const emailSchema = Joi.object<FormMetadataContactEmail>().keys({
-  address: emailAddressSchema,
+  address: emailAddressSchema.lowercase(),
```
colinrotherham: Loving all this @mokhld

Is there another ticket for when the email is typed it at file download time?

Presume we need to lower it there too so it matches

https://github.com/DEFRA/forms-designer/blob/f688767033d434978158bb800154bdf5b39f4b18/designer/server/src/routes/file/file.js#L155-L157

mokhld: Thanks for the heads up @colinrotherham 👌 . Haven't actually run the app as don't have the local dev account so I'm probably missing this and other things.

Question though, if we're normalising the string to lowercase upon save, then when we retrieve it, it will already be in lowercase?

mokhld: Sorry ignore me. Reading the file, I'm assuming this is we the user enters their email to download the file 😅

colinrotherham: Yeah, some people might like entering in **Firstname.Lastname** format 😮

PR #594, opened 2024-11-12
Conditions Break when retrospective changes are made (428569)

davidjamesstone:
The condition display text is synced to changes in the bound component and section
image

Remove "Delete" button and show message if the list is referenced
image

If the list item is referenced, the "Delete" button is removed and message displayed. Also, list item value and text is not editable.
image


designer/client/src/list/ListItemEdit.test.tsx
```diff
@@ -148,4 +148,22 @@ describe('ListItemEdit', () => {
...
+        data={data}
+        initialName="myList"
+        initialItemText="text a"
+      >
+        <ListItemEdit />
+      </RenderListEditorWithContext>
+    )
+
+    const $itemDeleteMessage = screen.getByText(
+      'This list item cannot be deleted as it is referenced in: "my condition"'
```
colinrotherham, line 164: Hey Dave, great work. Can we clarify what `"my condition"` is?

Given another name, it's not clear what references the list item (could be a component?)

```suggestion
      'This list item cannot be deleted as it is referenced by condition: "my condition"'
```

designer/client/src/data/condition/getConditions.ts
```diff
@@ -0,0 +1,32 @@
...
+import { type FieldDef } from '~/src/data/component/fields'
+
+/**
+ * @param data - the form definition data
+ * @param fields - an array of current fields
+ */
+export function getConditions(data: FormDefinition, fields: FieldDef[]) {
+  const { conditions } = data
+
+  conditions.forEach((condition: ConditionWrapper) => {
```
colinrotherham, line 17: Sorry @davidjamesstone, can we fix this so `data` isn't modified?

Currently the call to `getConditions(data)` has side effects on the original `data` state

This has caused us bugs with React state getting out of sync, skipping renders or other oddities

```js
const conditions = getConditions(data)
console.log(data) // Original has been modified 🚨
```

E.g. See `updateLinksTo()` where the original `data` is returned when changes are unnecessary:

https://github.com/DEFRA/forms-designer/blob/ff8700c6e0a4165caccc4ed66e48a5717dc5f317/designer/client/src/data/page/updateLinksTo.ts#L8-L16

davidjamesstone: Resolved with ab09bf2b51ff968f8a33d0f15a19cfdfb250bc34

designer/client/src/list/ListItemEdit.test.tsx
davidjamesstone: Resolved with b06df58be0ff0f812620f34e625283411e7d2106

designer/client/src/Designer.tsx
```diff
@@ -31,7 +32,7 @@ function useData(props: Readonly<Props>): DataContextType {
    */
   const get = useCallback(async () => {
     const updated = await form.get(props.meta.id)
```
colinrotherham, line 34: Watch out here, the patched data goes to `setState()` but the unpatched data is returned

Can we make sure the cloned data from `updateConditions()` goes to both?

```suggestion
    const updated = updateConditions(await form.get(props.meta.id))
```

davidjamesstone: Thanks, I missed that.  Resolved with ac0723e4e3f15eb77c049dff5c8274171b0ce4d0

designer/client/src/list/ListEdit.test.tsx
```diff
@@ -41,9 +41,56 @@ describe('ListEdit', () => {
     const $listCaption = screen.getByText('List items')
     const $listHint = screen.getByText('Enter a unique name for your list')
     const $listLink = screen.getByText('Add a new list item')
+    const $listDelete = screen.getByText('Delete')
```
colinrotherham, line 44: Think we're alright, but these old `getByText()` tests sometimes matched text not form elements

```suggestion
    const $buttonDelete = screen.getByRole('button', { name: 'Delete' })
```

designer/client/src/Designer.tsx
```diff
@@ -31,16 +32,21 @@ function useData(props: Readonly<Props>): DataContextType {
...
   }, [props])
 
   /**
    * Save form definition
    */
   const save = useCallback(
     async (updated: FormDefinition) => {
-      await form.save(props.meta.id, updateStartPage(updated))
+      const definition = updateConditions(updateStartPage(updated))
+      await form.save(props.meta.id, definition)
```
colinrotherham, line 48: @davidjamesstone If we're already fixing the data on `get()` do we need to fix it again on `save()`?

davidjamesstone: Without it the saved state (DB) could drift/not be as up to date as it could be.

`get()` -> `updateConditions` -> `User updates a section title` -> `save()`

We need to call `updateConditions` before save else the changes to the section title will not be reflected in the condition display text (in the DB).

If we don't mind the drift, the fix up before `save` can be removed.

designer/client/src/list/ListEdit.test.tsx
davidjamesstone: Resolved with 02835de15b88969cc9b12964dc5e0a45916d6d6b

designer/client/src/data/condition/getConditions.ts
colinrotherham: Sorry to come back to this one

I've had a look at [`updateConditions()` in `ab09bf2`](https://github.com/DEFRA/forms-designer/commit/ab09bf2b51ff968f8a33d0f15a19cfdfb250bc34#diff-1cc829d1a36951e1a446bc0eb972b5cb846e7988e4103c9f6236cd2781a43af6R14-R18) and we're cloning the definition every time it's called, even when no fixes are required. Are you sure we can't address this?

E.g. Here's the test for `updateStartPage()` where we use `toStrictEqual()` to confirm the object was left untouched when no start page changes were necessary:

https://github.com/DEFRA/forms-designer/blob/055dd0c8451eb3d522aa1e21d865d06818beb35c/model/src/form/utils/update-start-page.test.ts#L139-L145

davidjamesstone: Resolved with 8570a50eb10ad66109040f811c36e5284c143d75. We have to walk the tree to see if anything would change, if it doesn't return the data as-is, otherwise clone and alter.

designer/client/src/data/condition/updateConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // Determine if any of the conditions need fixing
+  const hasConditionsChanged = data.conditions.some((wrapper) => {
+    return wrapper.value.conditions.some((condition) => {
+      if (hasConditionField(condition)) {
+        return checkCondition(condition)
+      } else if (hasConditionGroup(condition)) {
+        condition.conditions
+          .filter((condition) => hasConditionField(condition))
+          .forEach(checkCondition)
```
colinrotherham, line 33: Have we definitely got coverage for condition groups?

The result of this filter isn't returned anywhere

designer/client/src/data/condition/updateConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+  // Fix up condition field display text
+  const fixUpCondition = (condition: ConditionData) => {
+    const field = checkCondition(condition)
+
+    if (field) {
+      condition.field.display = field.label
+    }
+  }
+
+  conditions.forEach((condition: ConditionWrapper) => {
```
colinrotherham, line 57: I thought we had an ESLint rule for this but we can remove types when they're already known

```suggestion
  conditions.forEach((condition) => {
```

designer/client/src/data/condition/updateConditions.ts
colinrotherham: Did you mean this?

```suggestion
        return condition.conditions
          .filter(hasConditionField)
          .some(checkCondition)
```

colinrotherham: Perhaps we move it to new `hasOutdatedCondition()` similar to `hasNestedCondition()`?

With some tests to make sure it correctly flags grouped-but-outdated conditions

https://github.com/DEFRA/forms-designer/blob/8570a50eb10ad66109040f811c36e5284c143d75/model/src/conditions/helpers.ts#L41-L43

designer/client/src/data/condition/updateConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+  type ConditionWrapper,
+  type FormDefinition
+} from '@defra/forms-model'
+
+import { getFields } from '~/src/data/component/fields.js'
+
+/**
+ * Update conditions to keep display text in sync with the component and section titles
+ */
+export function updateConditions(data: FormDefinition) {
```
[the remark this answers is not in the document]
davidjamesstone: Resolved with a4660973764cc306d010ae8114793c46d5ee1ff0 and 8813bcaf4fe837b519f78e5af89cf4e8450dffd3

designer/client/src/data/condition/updateConditions.ts
davidjamesstone: I think I'll keep it as-is for now. The function currently [depends](https://github.com/DEFRA/forms-designer/pull/594/files#diff-c4a25ac12bca23bbac8ac6afcf8c74955f56d66932edb22e1129fa4ff473b6a1R14) on the `FieldDef` class which currently lives in `designer/client/src/data/component/fields.ts`.

If an `hasOutdatedCondition` type function ever has any use outside this `fixupConditions` utiity function, let's move it then.

colinrotherham: No worries

>Have we definitely got coverage for condition groups?
>
>The result of this filter isn't returned anywhere

Can you add tests for groups though?

Just to confirm https://github.com/DEFRA/forms-designer/pull/594/commits/ee9f75717b3f9e5683af82927dae3da4e310999f works as expected

davidjamesstone: Resolved with a73ab8aa5eb4296f92b5226e6b57497a1c48f149

colinrotherham (approved):

PR #603, opened 2024-11-19
Escape single quotes in condition values when building the expression

davidjamesstone:
Condition values that contain any single quotes would cause the form definition to fail to load in `forms-runner`.
This is because the condition expression itself is [wrapped in single quotes](https://github.com/DEFRA/forms-designer/blob/e4230d468acd3187256cc806787ef4d2daf2db04/model/src/conditions/condition-operators.ts#L166). This causes an exception when the expression is subsequently [passed into](https://github.com/DEFRA/forms-runner/blob/main/src/server/plugins/engine/models/FormModel.ts#L154) `expr-eval` inside `forms-runner`.

This PR escapes any single quotes found in the condition values when building the expression.

### DIFF
--- a/model/src/conditions/condition-operators.ts
+++ b/model/src/conditions/condition-operators.ts
@@ -163,7 +163,7 @@ function formatValue(
     return value.toExpression()
   }
 
-  return `'${value.toExpression()}'`
+  return `'${value.toExpression().replace(/'/g, "\\'")}'`
 }
 
 export const absoluteDateOperatorNames = Object.keys(
--- a/model/src/conditions/condition-operators.ts
+++ b/model/src/conditions/condition-operators.ts
@@ -160,10 +160,10 @@ function formatValue(
     field.type === ComponentType.NumberField ||
     field.type === ComponentType.YesNoField
   ) {
-    return value.toExpression()
+    return value.toValue()
   }
 
-  return `'${value.toExpression().replace(/'/g, "\\'")}'`
+  return `'${value.toValue().replace(/'/g, "\\'")}'`
 }
 
 export const absoluteDateOperatorNames = Object.keys(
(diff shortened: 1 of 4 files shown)


mokhld (approved):

model/src/conditions/condition-operators.ts
```diff
@@ -163,7 +163,7 @@ function formatValue(
     return value.toExpression()
   }
 
-  return `'${value.toExpression()}'`
+  return `'${value.toExpression().replace(/'/g, "\\'")}'`
```
colinrotherham, line 166: This PR looks amazing 🙌

Two questions though, shouldn't it be in [`toExpression()`](https://github.com/DEFRA/forms-designer/blob/51948c8d2179c47c0b78fdfb6aafc686172f93d4/model/src/conditions/condition-values.ts#L34-L36) where we add the patch?

```patch
  toExpression() {
-   return this.value
+  return this.value.replace(/'/g, "\\'")
  }
```

And secondly, do we need to escape other JavaScript characters like `[`, `{`?

davidjamesstone:
```
function formatValue(
  field: Pick<ComponentDef, 'type'>,
  value: Condition['value']
) {
  if (
    (field.type === ComponentType.DatePartsField &&
      value.type === ConditionType.RelativeDate) ||
    field.type === ComponentType.NumberField ||
    field.type === ComponentType.YesNoField
  ) {
    return value.toExpression()
  }

  return `'${value.toExpression().replace(/'/g, "\\'")}'`
}
```

The issue is less about escaping any character but ensuring we aren't generating an invalid string. In the last line of that code block, we're wrapping the value expression itself in single quotes, therefore any single quotes inside the string need excaping.

Also, we could do this inside `toExpression` but it's only here where it's required (we don't want escaping for `DatePartsField`, `RelativeDate`, `NumberField`, `YesNoField` as they don't get wrapped in single quotes)

colinrotherham: Yeah, I'd expect `toExpression()` to always be a valid expression

Happy to move things

PR #604, opened 2024-11-20
Fix compiler issue with mixed component types

colinrotherham:
This PR fixes a minor type issue when mixing form + content components

All base interfaces `FormFieldBase`, `ContentFieldBase` and `DateFieldBase` now have properties in common

Preventing the following error:

```console
  Types of property 'pages' are incompatible.
    Type '({ path: string; components: { type: ComponentType.YesNoField; name: string; title: string; options: { required: true; }; }[]; section: string; next: ({ path: string; condition?: undefined; } | { path: string; condition: string; })[]; title: string; controller?: undefined; } | ... 4 more ... | { ...; })[]' is not assignable to type 'Page[]'.
```

### DIFF
--- a/model/src/components/types.ts
+++ b/model/src/components/types.ts
@@ -48,17 +48,18 @@ interface ContentFieldBase {
     | ComponentType.List
   name: string
   title: string
+  options?: {
+    required?: undefined
+    optionalText?: undefined
+  }
 }
 
-interface DateFieldBase {
+interface DateFieldBase extends FormFieldBase {
   type: ComponentType.DatePartsField | ComponentType.MonthYearField
   name: string
   title: string
   hint?: string
-  options: {
-    required?: boolean
-    optionalText?: boolean
-    classes?: string
+  options: FormFieldBase['options'] & {
     maxDaysInPast?: number
     maxDaysInFuture?: number
   }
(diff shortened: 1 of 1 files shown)


model/src/components/types.ts
```diff
@@ -48,17 +48,18 @@ interface ContentFieldBase {
     | ComponentType.List
   name: string
   title: string
+  options?: {
+    required?: undefined
+    optionalText?: undefined
```
mokhld, line 53: @colinrotherham can we not just remove these properties if they're set to `undefined` or explicitly use `never` (which is slightly different and assumes that they'll never be used so the linter will always pick this up I think)?

colinrotherham: It's the very thing that fixes the issue 😔

The compiler sees `FormFieldBase`, `ContentFieldBase` and `DateFieldBase` all with common properties

We had to do the same with [page types that don't support sections](https://github.com/search?q=repo%3ADEFRA%2Fforms-designer+%22section%3F%3A+undefined%22&type=code)

colinrotherham: Forgot to say, using `never` is perfect for clarity but causes more `ComponentDef[]` union problems

I've not found anything better than `?: undefined` to let us simplify checks like this:

```patch
- if ('required' in options && options.required) {
+ if (options.required) {
    // …
  }
```

mokhld: Ah in that case it makes sense. Never the 'best' option ofc, but it does simplify things. If you were doing a larger refactor I would've said that having an interface of `FormOptions` with the common options (there's quite a few reused) then we could do something like this:
```
type SomeUndefinedOptions = {
  [K in keyof FormOptions]?: undefined
}
```
But tbf that's not too explicit either. _sigh_ The joys of Typescript.

mokhld (approved):

PR #613, opened 2024-11-29
feature(368034): Added pagination and query shared types and validation

mokhld:
These are shared types that will be used in https://github.com/DEFRA/forms-manager/pull/379.


model/src/common/pagination/types.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @property {number} perPage - The number of items per page.
+ * @property {number} totalItems - The total number of items available.
+ * @property {number} totalPages - The total number of pages available.
+ */
+export interface PaginationResult {
+  page: number
+  perPage: number
+  totalItems: number
+  totalPages: number
+}
```
colinrotherham, line 13: PR looks ace

Would you mind moving the comments above the properties?

Just to make sure type tooltips get the descriptions coming through like this:

Descriptions in tooltips

colinrotherham: Which I know is a pain, because they started life as JSDoc 😣

Would welcome a tech debt PR one day to migrate `model` to JSDoc (with generated types still)

model/src/index.ts
```diff
@@ -7,8 +7,10 @@ export * from '~/src/conditions/index.js'
 export * from '~/src/pages/index.js'
 export * from '~/src/utils/helpers.js'
 export * from '~/src/utils/markdown.js'
+export * from '~/src/common/pagination/index.js'
```
colinrotherham, line 10: Super non-blocker, but these exports are currently ordered A-Z

Maybe there's an ESLint option we could add (similar to [`import/order`](https://github.com/DEFRA/forms-designer/blob/main/.eslintrc.cjs#L95-L96))

model/src/common/pagination/types.ts
mokhld: I always like to be as descriptive as possible but it can be a bit uneeded tbf.

model/src/index.ts
mokhld: Really didn't spot that haha. I think this would make a nice-to-have tech debt ticket...

model/src/common/pagination/types.ts
colinrotherham: Ah sorry @mokhld yeah same

I meant like this. When the comment is above the property it appears in tooltips

```ts
/**
 * Result of pagination containing page information
 */
export interface PaginationResult {
  /**
   * The current page number.
   */
  page: number

  /**
   * The number of items per page.
   */
  perPage: number

  /**
   * The total number of items available.
   */
  totalItems: number

  /**
   * The total number of pages available.
   */
  totalPages: number
}
```

mokhld: ahh got you.

model/src/common/pagination/types.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface Meta {
+  /**
+   * The pagination details
+   */
+  pagination?: PaginationResult
+}
+
+/**
+ * Generic result type for paginated queries
+ * @template T The type of items in the result
```
colinrotherham, line 46: Left over from JSDoc?

```suggestion
```

model/src/common/pagination/types.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+export interface QueryResult<T> {
```
colinrotherham, line 48: Then probably best to name this `T` as we know what it is already

```suggestion
export interface QueryResult<ItemType> {
```

model/src/common/pagination/types.ts
mokhld: indeed it was 🥴

model/src/common/pagination/types.ts
mokhld: I like `T` :/ , but agreed. I've renamed it to `Model` though to represent the mapped model from the db.

mokhld:
> Thanks for doing this @mokhld
> 
> Is everything working when you do `npm link`?
> 
> Last of all, it's worth checking with the team on naming conventions for:
> 
> ```
> page
> perPage
> totalItems
> totalPages
> ```
> 
> Check everyone's happy
> 
> Code wise all good though ✅

Thanks @colinrotherham . To answer your question, yep all good with `npm link`- it can sometimes be a bit fiddly, but so far has been stable.
I think the guys have seen the `form-manager` PR but will double check with them all 👍

PR #616, opened 2024-12-05
Feature 368034 : Refresh- Pagination for /library page

mokhld:
### Ticket Link
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/368034

### Description
This PR introduces pagination for the `/library` homepage in forms-designer. It handles building the pagination 'server-side' and thus abstracts the complexity away from the nunjucks template view model.
It also adds the pagination component with tests, allowing it to be re-used anywhere else that requires pagination quite easily.

### Flow diagram for `buildPaginationPages()` method
This was drawn up during writing this method as I find it helps to visually map out logic/ complexities.

buildPaginationPages() flow diagram


designer/server/src/lib/forms.test.js
```diff
@@ -0,0 +1,521 @@
...
+   */
+  const formDefinition = {
+    name: 'Test form',
+    pages: [],
+    conditions: [],
+    sections: [],
+    lists: []
+  }
+
+  describe('Forms library routes (mocked forms functions)', () => {
```
mokhld, line 66: Not ideal, but as we were mocking the entire `/forms.js` file, I had to restructure the file with the mocked methods and a separate describe block for the actual methods to test from the file. It probably would've been easier to create a separate sheet in hindsight, but naming might be confusing.

designer/server/src/lib/forms.test.js
```diff
@@ -0,0 +1,521 @@
...
+        expect(fetch.getJson).toHaveBeenCalledWith(expect.any(URL), {
+          headers: { Authorization: `Bearer ${token}` }
+        })
+
+        const fetchGetJsonMock = /** @type {jest.Mock} */ (fetch.getJson)
+
+        /** @type {Array<[URL, object]>} */
+        const mockCalls = fetchGetJsonMock.mock.calls
+
+        const calledUrl = /** @type {URL} */ (mockCalls[0][0])
```
mokhld, line 443: Not ideal, but due to type checking issues when accessing mock.calls on the mocked fetch.getJson func

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
...
+        expect($nextLink).toBeInTheDocument()
+        expect($nextLink).toHaveAttribute('href', '/library?page=3&perPage=10')
+        expect($nextLink).toHaveTextContent('Next')
+      })
+
+      it('should render the correct page numbers', () => {
+        const $pages = $pagination?.querySelectorAll('.govuk-pagination__item')
+        expect($pages).toHaveLength(5)
+
+        expect($pages?.[0].textContent?.trim()).toBe('1')
```
colinrotherham, line 76: Love seing all these Do the DOM assertions do the trim bit for us?

```suggestion
        expect($pages?.[0]).toHaveTextContent('1')
```

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
...
+          'govuk-pagination__item--current'
+        )
+
+        expect($pages?.[3].textContent?.trim()).toBe('⋯') // horizontal ellipsis!
+        expect($pages?.[3].classList).not.toContain(
+          'govuk-pagination__item--current'
+        )
+
+        expect($pages?.[4].textContent?.trim()).toBe('5')
+        expect($pages?.[4].classList).not.toContain(
```
colinrotherham, line 97: Is it a bit nicer with `toHaveClass()`?

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
...
+                  number: '5',
+                  href: '/library?page=5&perPage=10',
+                  current: false
+                }
+              ]
+            }
+          }
+        })
+
+        $pagination = document.querySelector('.govuk-pagination')
```
colinrotherham, line 45: Any chance we can use the accessibility-aware [`ByRole` queries](https://testing-library.com/docs/queries/byrole/)?

```suggestion
        $pagination = container.getByRole('navigation', {
          name: 'Pagination'
        })
```

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
+import { renderMacro } from '~/test/helpers/component-helpers.js'
+
+describe('Pagination Component', () => {
+  /** @type {Element | null} */
+  let $pagination = null
```
colinrotherham, line 5: If we use role queries ([see comment](https://github.com/DEFRA/forms-designer/pull/616/files#r1875775788)), we don't need to handle `null` 💪 

```suggestion
  /** @type {HTMLElement} */
  let $pagination
```

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
...
+      })
+
+      it('should render the pagination component', () => {
+        expect($pagination).toBeInTheDocument()
+      })
+
+      it('should have the correct previous and next links', () => {
+        const $previousLink = $pagination?.querySelector(
+          '.govuk-pagination__prev a'
+        )
```
colinrotherham, line 55: You might have seen us use `within()` too?

```js
import { within } from '@testing-library/dom'
```

Let's you take any HTML element and query against it again using Testing Library

```suggestion
        const $previousLink = within($pagination).getByRole('link', {
          name: 'Previous',
        })
```

designer/server/src/common/components/pagination/template.test.js
mokhld: Not sure if we can do this, as there might be times when pagination isn't loaded at all e.g. only one page

colinrotherham: That's fair. We can still use the role queries though? Just with the `| null` type back in

mokhld: Managed to find a workaround for this, so we can keep it as you suggested 👍

designer/server/src/common/components/pagination/template.test.js
```diff
@@ -0,0 +1,580 @@
...
+        })
+
+        $pagination = document.querySelector('.govuk-pagination')
+      })
+
+      it('should not render the previous link', () => {
+        const $previousLink = $pagination?.querySelector(
+          '.govuk-pagination__prev a'
+        )
+        expect($previousLink).toBeNull()
```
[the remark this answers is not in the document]
mokhld: Yeah brilliant, that's been updated.

designer/server/src/common/components/pagination/template.test.js
mokhld: It is indeed!

designer/server/src/routes/forms/library.js
```diff
@@ -11,14 +12,39 @@ export default [
...
-      const model = await library.listViewModel(token)
-
-      return h.view('forms/library', model)
-    },
     options: {
+      handler: async (request, h) => {
+        const { auth, query } = request
+        const token = auth.credentials.token
+
+        const page = Number(query.page) || 1
```
davidjamesstone, line 20: Can we use Joi with defaults for this like we do in forms-manager?
`options.validate.query = joi.object().keys({ page: joi.number().default(1), ... })`.

designer/server/src/routes/forms/library.js
```diff
@@ -11,14 +12,39 @@ export default [
...
+            // Redirect to the first page
+            const redirectUrl = new URL('/library', config.appBaseUrl)
+            redirectUrl.searchParams.set('page', '1')
+            redirectUrl.searchParams.set('perPage', String(perPage))
+            return h.redirect(redirectUrl.pathname + redirectUrl.search)
+          }
+        }
+
+        // Redirect to include pagination params if missing
+        if (!('page' in query) && !('perPage' in query) && model.pagination) {
```
davidjamesstone, line 39: I think this can be removed if you use joi on the route options.

designer/server/src/common/components/pagination/template.njk
```diff
@@ -0,0 +1,27 @@
+{% from "govuk/components/pagination/macro.njk" import govukPagination %}
+
+{% if params.pagination and params.pagination.totalPages > 1 %}
+  {% set previous = null %}
+  {% if params.pagination.page > 1 %}
+    {% set previous = {
+      href: (params.baseUrl | default('')) ~ "?page=" ~ (params.pagination.page - 1) ~ "&perPage=" ~ params.pagination.perPage,
```
davidjamesstone, line 7: What are these tildes, whitespace control? I can't see them documented in njk.

mokhld: For string concatenation. So for example: 
```
{% set firstName = "John" %}
{% set lastName = "Smith" %}

{# Using tilde for concatenation #}
{{ firstName ~ " " ~ lastName }}

{# Makes the ooutput: "John Smith" #}
```

designer/server/src/models/forms/library.js
```diff
@@ -8,20 +8,120 @@ import {
...
  */
-export async function listViewModel(token) {
+export async function listViewModel(token, paginationOptions) {
   const pageTitle = 'Forms library'
-  const formItems = await forms.list(token)
+
+  const formResponse = await forms.list(token, paginationOptions)
+
+  const formItems = formResponse.data
+  const paginationMeta = formResponse.meta.pagination ?? null
```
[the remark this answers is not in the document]
mokhld: Yeah absolutely, good shout.

davidjamesstone (approved):

PR #627, opened 2024-12-12
Add error page for HTTP 400

alexluckett:
Supports 368034 to present a friendly error rather than JSON if the user changes the URL parameters.

### DIFF
--- a/designer/server/src/views/400.njk
+++ b/designer/server/src/views/400.njk
@@ -5,7 +5,7 @@
   {% call appPageBody({
     heading: pageHeading
   }) %}
-    <p class="govuk-body">The service couldn't process your request.</p>
+    <p class="govuk-body">The service could not process your request.</p>
     <p class="govuk-body"><a href="mailto:defraforms@defra.gov.uk" class="govuk-link">Contact the Defra forms team</a> to report this error.</p>
   {% endcall %}
 {% endblock %}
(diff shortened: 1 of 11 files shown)


designer/server/src/views/400.njk
```diff
@@ -0,0 +1,11 @@
+{% extends 'layouts/page.njk' %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
```
colinrotherham, line 2: Slowly trying to switch to double quotes everywhere in Nunjucks

```suggestion
{% extends "layouts/page.njk" %}

{% from "govuk/components/button/macro.njk" import govukButton %}
```

designer/server/src/views/400.njk
```diff
@@ -0,0 +1,11 @@
...
+
+{% block content %}
+  {% call appPageBody({
+    heading: pageHeading
+  }) %}
+    <p class="govuk-body">The service couldn't process your request.</p>
```
colinrotherham, line 8: What does "process your request" mean to a user?

Watch out for the [negative contraction](https://www.gov.uk/guidance/style-guide/a-to-z#contractions) "couldn't" too

```suggestion
    <p class="govuk-body">The service could not process your request.</p>
```

designer/server/src/views/400.njk
alexluckett: Done for a few pages I found

davidjamesstone (approved):

designer/server/src/views/400.njk
alexluckett: I've put the question to Dan, will look to merge this unless they weigh in and let them improve it later.

PR #641, opened 2024-12-23
Feature 479075: Search (title) shared types and schema

mokhld:
## Story Link
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/479075

## Description
This PR introduces the shared types and schemas needed for the search addition for `GET /forms` in forms-manager.


model/src/common/search/index.ts
```diff
@@ -0,0 +1,17 @@
+import Joi from 'joi'
+
+import { type SearchOptions } from '~/src/common/search/types.js'
+
+/**
+ * Field definitions for search options.
+ */
+export const searchOptionFields = {
+  title: Joi.string().trim().min(1).optional()
```
colinrotherham, line 9: I like seeing `.trim()`, reminds me we should do it in a lot more places

model/src/common/search/index.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/**
+ * Joi schema for `SearchOptions` interface
+ * @see {@link SearchOptions}
```
colinrotherham, line 14: Can stick the link in the description if you like?

It lets you do Markdown style link text `{@link SearchOptions | Link text here}` too

```suggestion
 * Joi schema for {@link SearchOptions} interface
```

model/src/common/search/index.ts
colinrotherham, line 9: Can I double check what "optional" means in a GET sense?

Think we also need `.allow('')` for when `?title=` is an empty string

colinrotherham (approved):

model/src/common/search/index.ts
mokhld: That's more concise for sure.

model/src/common/search/index.ts
mokhld: That's a good spot! Joi wouldn't let an empty string through- thought it would as it's optional

mokhld: I think we can just remove `min(1)`

mokhld: Nope I was wrong 😅

PR #651, opened 2025-01-06
Allow deletion of a form if not live

alexluckett:
- Add secondary links to the overview page form management actions
- Allows a form to be deleted from the overview page (with a confirmation page) if there is no live form
- Shows a confirmation page on the library when the form has been successfully deleted


alexluckett:
This is pretty much done, just waiting for finalised designs regarding the notification banner in the library.

davidjamesstone (approved):

PR #663, opened 2025-01-08
Feature/479075 search title

mokhld:


designer/server/src/common/components/search/_search.scss
```diff
@@ -0,0 +1,26 @@
+@use "govuk-frontend/govuk/base" as *;
```
colinrotherham, line 1: I'll give you a nudge about this @mokhld

You've not hit it yet, but you'll notice `$from: wide` is missing in how we currently import

Will open another PR to resolve it

colinrotherham: Ready for review in:

* https://github.com/DEFRA/forms-designer/pull/666

Once merged, there's a new `@forward` Sass partial to import:

```suggestion
@use "../govuk-frontend" as *;
```

mokhld: Thanks for the heads-up, @colinrotherham! I'll add that in once your PR is merged.

colinrotherham: No problem at all, merged 🙌

alexluckett (approved):
🚀

davidjamesstone (approved):

PR #687, opened 2025-01-28
Disallow concurrent sessions

alexluckett:
Blocks duplicate sessions by identifying a duplicate (based on a unique `flowId` per login flow instance). When a duplicate session is identified, the session is rejected and the user is bounced back to the usual callback route. This will then sign the user out of AAD.


mokhld (approved):
LGTM 🚀

PR #689, opened 2025-01-29
Chore 494884: Escape html

mokhld:
## Ticket link
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/494884

Example:

Screenshot 2025-01-29 at 15 04 46

### DIFF
--- a/designer/server/src/views/forms/overview.njk
+++ b/designer/server/src/views/forms/overview.njk
@@ -74,7 +74,7 @@
   {%- set phoneHtml %}
     {% if form.contact.phone %}
       <div class="app-prose-scope">
-        {{ form.contact.phone | markdown | escape | safe }}
+        {{ form.contact.phone | markdown | safe }}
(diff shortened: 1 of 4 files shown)


davidjamesstone (approved):

designer/server/src/views/forms/overview.njk
```diff
@@ -74,7 +74,7 @@
   {%- set phoneHtml %}
     {% if form.contact.phone %}
       <div class="app-prose-scope">
-        {{ form.contact.phone | markdown | safe }}
+        {{ form.contact.phone | markdown | escape | safe }}
```
alexluckett, line 77: Markdown is meant to be rendered here, but just not HTML injected via user input.

So it might need to be `| escape | markdown | safe` instead, with the order switched.

Ditto for the other ones below.

alexluckett (approved):

PR #719, opened 2025-02-26
466796 enhancements and 468287 implementation

jbarnsley10:
Allows a user to edit a page
Allows a user to edit a question
Allows a user to add an additional question
Eliminated all lint warnings on new code
Allows a user to preview a page or a form
Added no-visited-state class to action links

### DIFF
--- a/designer/server/src/models/forms/editor-v2/questions.js
+++ b/designer/server/src/models/forms/editor-v2/questions.js
@@ -12,7 +12,7 @@ import {
   baseModelFields,
   getFormSpecificNavigation
 } from '~/src/models/forms/editor-v2/common.js'
-import { formOverviewPath } from '~/src/models/links.js'
+import { editorv2Path, formOverviewPath } from '~/src/models/links.js'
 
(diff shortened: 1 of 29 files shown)


designer/server/src/models/forms/editor-v2/questions.js
```diff
@@ -122,12 +120,15 @@ export function questionsViewModel(
     ? formValues?.guidanceText
     : guidanceTextFallback
 
+  const baseUrl = `${editorv2Path(metadata.slug)}page/${pageId}`
```
alexluckett, line 123: missing `/` after `editorV2Path(...)`

jbarnsley10: editorV2Path actually leaves a trailing slash already if you don't supply the 'page' param. I could pass 'page/${pageId}' to the editorV2Path() to achieve the same result, but the code above is correct as it stands I think

alexluckett (approved):

PR #720, opened 2025-02-28
Feat/468274 page titles

jbarnsley10:
Reworked page title logic to simplify things
In V2 page schema, page title can be a blank string and forms-runner determines whether to display page title from 'page title' value or first question title
This PR needs to merge in tandem with the equivalent model changes for forms-manager and forms-runner

### DIFF
--- a/model/src/form/form-definition/index.ts
+++ b/model/src/form/form-definition/index.ts
@@ -200,6 +200,29 @@ export const pageSchema = Joi.object<Page>().keys({
   view: Joi.string().optional()
 })
 
+/**
+ * V2 engine schema - used with new editor
+ * `/status` is a special route for providing a user's application status.
+ *  It should not be configured via the designer.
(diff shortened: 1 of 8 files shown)


model/src/form/form-definition/index.ts
```diff
@@ -276,7 +299,11 @@ export const formDefinitionSchema = Joi.object<FormDefinition>()
     startPage: Joi.string().optional(),
     pages: Joi.array<Page>()
       .required()
-      .items(pageSchema)
+      .when('engine', {
+        is: 'V2',
+        then: Joi.array<Page>().items(pageSchemaV2),
+        otherwise: Joi.array<Page>().items(pageSchema)
+      })
```
alexluckett, line 306: do we need an entire new page schema for this, or can we just set the page title to be optional within the existing schema?

PR #729, opened 2025-03-05
Automatically select the AAD account to sign the user out of

alexluckett:
The user may have multiple accounts in AAD. This automatically selects the right account when signing the user out, so they don't need to pick and accidentally sign themselves out of the wrong account.

This PR replaces the "force sign out" flag we previously had, instead opting for an optional override of the `logoutHint` query parameter. If it's provided (e.g. for a force signout), it's used else it falls back on the one from the session.

This closes off story 501146.

### DIFF
--- a/designer/package.json
+++ b/designer/package.json
@@ -79,6 +79,7 @@
     "i18next-http-backend": "^3.0.1",
     "ioredis": "^5.4.2",
     "joi": "^17.13.3",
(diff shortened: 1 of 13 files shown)


alexluckett:
Somehow broke the tests on my last commit....

davidjamesstone (approved):

PR #730, opened 2025-03-06
Added Markdown component

jbarnsley10:
Added Markdown component to be able to differentiate between Html component and Markdown component
Amended page settings and summary declaration to use Markdown component
NOTE:
1. The Markdown component is available in the old editor
2. I have not updated the forms-runner to handle it yet
3. Currently I have diverted from the screen design by using the term 'Markdown' rather than 'Declaration' in the pages overview. This allows the summary page to be more generic, and potentially have more than one markdown


designer/server/src/__stubs__/form-definition.js
```diff
@@ -255,7 +257,7 @@ export const testFormDefinitionWithExistingGuidance = {
       components: [
         {
           id: '45678',
-          type: ComponentType.Html,
+          type: ComponentType.Markdown,
```
alexluckett, line 260: accident?

mokhld (approved):

davidjamesstone:
This is great Jez :rocket:

PR #743, opened 2025-03-11
Fix 518654: Prevent special characters in component names for condition expressions

mokhld:
## Ticket
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/518654

## Problem
Component names containing special characters (like dashes, periods, etc.) were causing condition evaluations to fail. This happened because expr-eval misinterprets these characters as mathematical operators. For example, a component name like "hello-world" is interpreted as "hello minus world".

## Solution
A preventative approach by adding Joi validation to restrict component names to only alphanumeric characters and underscores, thus preventing the problem at the source:

```typescript
name: Joi.when('type', {
  is: Joi.string().valid(
    ComponentType.Details,
    ComponentType.Html,
    ComponentType.InsetText,
    ComponentType.Markdown
  ),
  then: Joi.string()
    .pattern(/^\w+$/)
    .optional(),
  otherwise: Joi.string()
    .pattern(/^\w+$/)
})
```

Essentially ensuring that component names can only contain letters, numbers, and underscores across all component types, preventing expr-eval issues before they arise as an exception will be thrown.

## Alternative Approach
As an alternative, I explored handling special characters using a transformation approach in the forms-runner (see: [DEFRA/forms-runner PR #744](https://github.com/DEFRA/forms-runner/pull/744)) which uses a `safeFieldName` helper function (see commit: https://github.com/DEFRA/forms-designer/pull/743/commits/3a38afe7127ddea334cd155b624ecfd964aa3fcf) to:

1. Transform component names with special characters into safe variable names
2. Create aliases in the evaluation context that contain both the original and transformed names

For example:
- Component name: `hello-world` 
- Becomes in expressions: `_field_hello_world`
- Both variables exist in context: `{"hello-world": "value", "_field_hello_world": "value"}`

Although this approach works and maintains backward compatibility, we've decided that preventing the issue upfront with validation is cleaner and more deterministic however would ultimately require developers using the `name` property to change it.


davidjamesstone (approved):
LGTM Mo.

One thought is if `[a-zA-Z_]+` is better than `\w+`, in case the component name starts with, or is only, digits?.

E.g. `name: "1"` or `name: "1foo"` - does these fail?

mokhld:
> LGTM Mo.
> 
> One thought is if `[a-zA-Z_]+` is better than `\w+`, in case the component name starts with, or is only, digits)?.
> 
> E.g. `name: "1"` or `name: "1foo"` - do these fail?

Thanks @davidjamesstone . I've added two tests for these scenarios and they're passing.

PR #744, opened 2025-03-12
Feat/481464  migrate on homepage

whitewaterdesign:
## Migrate v1 definition to v2 

When a user calls the v2 homepage (/library/v2-form/editor-v2/pages) on a v1 draft form -  a call is sent through to migrate the form to v2.  The page is then opened with the updated definition.

### DIFF
--- a/designer/server/src/routes/forms/editor-v2/pages.js
+++ b/designer/server/src/routes/forms/editor-v2/pages.js
@@ -1,5 +1,8 @@
+import { Engine } from '@defra/forms-model'
+
 import * as scopes from '~/src/common/constants/scopes.js'
 import { sessionNames } from '~/src/common/constants/session-names.js'
+import * as editor from '~/src/lib/editor.js'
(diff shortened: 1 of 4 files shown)


davidjamesstone (approved):

designer/server/src/routes/forms/editor-v2/pages.js
```diff
@@ -21,7 +24,13 @@ export default [
       const { slug } = params
 
       const metadata = await forms.get(slug, token)
-      const definition = await forms.getDraftFormDefinition(metadata.id, token)
+      const formId = metadata.id
+
+      let definition = await forms.getDraftFormDefinition(formId, token)
+
+      if (definition.engine === Engine.V1) {
+        definition = await editor.migrateDefinitionToV2(formId, token)
```
alexluckett, line 32: for the future, we'll likely want a confirmation dialog before doing this but for now as it's a relatively simple change we're ok.

alexluckett (approved):

PR #746, opened 2025-03-12
Populates name with randomId

jbarnsley10:
Populates name with randomId
Saves shortDescription to shortDescription

### DIFF
--- a/designer/server/src/models/forms/editor-v2/question-details.js
+++ b/designer/server/src/models/forms/editor-v2/question-details.js
@@ -20,10 +20,11 @@ function mapToQuestionDetails(question) {
     return {}
   }
   return {
+    name: question.name,
     question: question.title,
     hintText: question.hint,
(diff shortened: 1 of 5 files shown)


designer/server/src/models/forms/editor-v2/question-details.test.js
```diff
@@ -81,9 +82,41 @@ describe('editor-v2 - question details model', () => {
       expect(res.fields).toEqual({})
     })
   })
+
+  describe('mapToQuestionDetails', () => {
+    test('should map if question undefined', () => {
+      const res = mapToQuestionDetails(undefined)
+      expect(res).toEqual({
+        name: expect.anything(),
```
whitewaterdesign, line 90: you can do expect.any(String) here, it's slightly more clear.  I had problems trying to stub crypto.randomUUID - wondering if it's possible here?  Would make a stronger test.

PR #748, opened 2025-03-13
Feat/extending advanced settings

jbarnsley10:
Allows easier extensibility - at least as a next stage
Could be refined further in terms of implementing inheritance
I wasn't sure whether it was worth combining the field and field type into a single field type and passing that in the list of fields or whether to keep with just a model function to determine field type


designer/server/src/models/forms/editor-v2/question-details-advanced-settings.js
```diff
@@ -0,0 +1,137 @@
...
+  return fieldType
+}
+
+/**
+ * @param {string} fieldName
+ * @param {FormEditor} formValues
+ * @param { ErrorDetails | undefined } formErrors
+ */
+export function buildField(fieldName, formValues, formErrors) {
+  let field
```
whitewaterdesign, line 35: Not an issue, but you probably don't need to set field here, you could just return the field {...} on each branch of the switch...would save 7 lines.  Unless it's a pain with typing

whitewaterdesign (approved):
Looks good, a couple of possible refactors to reduce number of lines a bit, but not a deal breaker.

whitewaterdesign (changes requested):
conflict needs sorting

whitewaterdesign (approved):
Looks good to me, a couple of possible refactors on there.

designer/server/src/models/forms/editor-v2/question-details-advanced-settings.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @param {string} fieldName
+ * @param {FormEditor} formValues
+ * @param { ErrorDetails | undefined } formErrors
+ * @returns {GovukField}
+ */
+export function buildField(fieldName, formValues, formErrors) {
+  switch (fieldName) {
+    case QuestionAdvancedSettings.MinLength:
```
alexluckett, line 33: might be cleaner as an object for when this gets longer?

```
// designer/server/src/models/forms/editor-v2/field-options.js
export default {
    [QuestionAdvancedSettings.MinLength]: {
        name: 'minLength',
        id: 'minLength',
        label: {
          text: 'Minimum character length (optional)',
          classes: GOVUK_LABEL__M
        },
        hint: {
          text: 'The minimum number of characters users can enter'
        },
        classes: 'govuk-input--width-3'
      }
}

// designer/server/src/models/forms/editor-v2/question-details-advanced-settings.js
import * as allFieldOptions from 'designer/server/src/models/forms/editor-v2/field-options.js'
export function buildField(fieldName, formValues, formErrors) {
    const fieldOptions = allFieldOptions[fieldName]
    return {
        ...fieldOptions,
        value: formValues.minLength,
        insertValidationErrors(formErrors?.minLength)
    }
}
```

maybe?

jbarnsley10: Yep. Possibly. Can change it

alexluckett (approved):

PR #766, opened 2025-03-24
Add list group sub-types

jbarnsley10:
Added sub-options radios for list type in question-type screen

### DIFF
--- a/designer/server/src/models/forms/editor-v2/question-type.js
+++ b/designer/server/src/models/forms/editor-v2/question-type.js
@@ -2,6 +2,7 @@ import { ComponentType } from '@defra/forms-model'
 
 import {
   QUESTION_TYPE_DATE_GROUP,
+  QUESTION_TYPE_LIST_GROUP,
   QUESTION_TYPE_WRITTEN_ANSWER_GROUP
 } from '~/src/common/constants/editor.js'
 import { buildErrorList } from '~/src/common/helpers/build-error-details.js'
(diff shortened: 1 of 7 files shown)


designer/server/src/models/forms/editor-v2/question-type.js
```diff
@@ -64,7 +65,7 @@ const questionTypeRadioItems = /** @type {FormEditorCheckbox[]} */ ([
   },
   {
     text: 'A list of options that users can choose from',
-    value: ComponentType.SelectField
+    value: QUESTION_TYPE_LIST_GROUP
```
whitewaterdesign, line 68: Just wondering why this is different from the current pattern?

jbarnsley10: The list option is now a group (not a single value). The group functionality wasn't implemented previously for this option (to save time and wasn't needed for the earlier stories)

model/src/form/form-editor/index.ts
```diff
@@ -20,7 +20,7 @@ export const questionTypeSchema = Joi.string()
     ComponentType.TelephoneNumberField,
     ComponentType.FileUploadField,
     ComponentType.EmailAddressField,
-    ComponentType.SelectField
+    'list-group'
```
whitewaterdesign, line 23: suggest enum here to keep consistency

whitewaterdesign (commented):
Looks good, just if we use enums, it's going to be easier to use these fields in future.

PR #771, opened 2025-03-24
Feat/472356 yes no

jbarnsley10:
Added Yes/No question type
Fixed precision min value
Fix maxFuture/maxPast min value


designer/server/src/models/forms/editor-v2/advanced-settings-fields.js
```diff
@@ -325,7 +325,7 @@ export const allSpecificSchemas = Joi.object().keys({
     '*': 'Maximum length must be a positive whole number'
   }),
   precision: questionDetailsFullSchema.precisionSchema.messages({
-    '*': 'Precision must be a positive whole number'
+    '*': 'Precision must be a whole number between zero and 5'
```
alexluckett, line 328: Is the mix of words and numbers intentional? Should it not be `zero and five` or `0 and 5`?

(I have no idea by the way, just checking)

jbarnsley10: Haha - not sure. That's for the 'content' people to decide. This is just really a placeholder for the actually type of error. I changed '0' to 'zero' as it can often look like a capital 'O' - but I see your point

alexluckett (approved):

PR #773, opened 2025-03-25
Feat/527770 confirm migration

jbarnsley10:
Handles confirmation screen (and subsequent migration call) to migrate a V1 form to V2 when trying to access the new editor with an old form definition
This PR is chained to PR https://github.com/DEFRA/forms-designer/pull/772


alexluckett:
@jbarnsley10 could you sync this up with main please, it'll make it a bit easier to diff - lots of 'delete page' content in here which I'm guessing is a result of that.

alexluckett (approved):

PR #779, opened 2025-03-27
Fix/529909 change file upload

jbarnsley10:
Allows File Upload question to be change to a differnet type of question, or from say a TextField to a FileUpload question - and handles the controller property on the page in the JSON

### DIFF
--- a/designer/server/src/lib/editor.js
+++ b/designer/server/src/lib/editor.js
@@ -104,10 +104,8 @@ export async function updateQuestion(
   // Determine if page controller should change
   const page = getPageFromDefinition(definition, pageId)
   const origControllerType = page?.controller
-    ? { controller: page.controller }
-    : {}
-  const newControllerType = getControllerType(questionDetails)
-  if (origControllerType.controller !== newControllerType.controller) {
+  const { controller: newControllerType } = getControllerType(questionDetails)
+  if (origControllerType !== newControllerType) {
     // Update page controller
     const pageHeadingRequestUrl = new URL(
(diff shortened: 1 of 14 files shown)


designer/server/src/lib/editor.js
```diff
@@ -87,17 +88,39 @@ export async function addQuestion(formId, token, pageId, questionDetails) {
...
+  const newControllerType = getControllerType(questionDetails)
+  if (origControllerType.controller !== newControllerType.controller) {
+    // Update page controller
+    const pageHeadingRequestUrl = new URL(
+      `./${formId}/definition/draft/pages/${pageId}`,
+      formsEndpoint
+    )
+    await patchJsonByType(pageHeadingRequestUrl, {
+      payload: {
+        controller: newControllerType.controller ?? null
```
alexluckett, line 118: Passing the strings around rather than wrapper objects is a little cleaner?

```
  const origControllerType = page?.controller
  const { controller: newControllerType } = getControllerType(questionDetails)
```

e.g.

```
  const origControllerType = page?.controller
  const { controller: newControllerType } = getControllerType(questionDetails)

  if (origControllerType !== newControllerType) {
    // Update page controller
    const pageHeadingRequestUrl = new URL(
      `./${formId}/definition/draft/pages/${pageId}`,
      formsEndpoint
    )
    await patchJsonByType(pageHeadingRequestUrl, {
      payload: {
        controller: newControllerType ?? null
```

jbarnsley10: Good point. Pushed a change

alexluckett (approved):

PR #787, opened 2025-04-02
Feat/481040 radio save

jbarnsley10:
Saves radio/checkbox lists to the server
- significant rework of session management to avoid collisions if user has multiple tabs open


designer/server/src/__stubs__/form-definition.js
```diff
@@ -506,6 +506,54 @@ export const testFormDefinitionWithFileUploadPage = {
...
+  pages: [
+    {
+      id: 'p1',
+      path: '/radio-question',
+      title: 'Radio question',
+      section: 'section',
+      components: [
+        {
+          id: 'q1',
+          type: ComponentType.RadiosField,
```
whitewaterdesign, line 523: would it be worth adding a factory for radio components?

alexluckett (commented):
The "Cancel" button when adding a list item doesn't preserve any change to made to the form, e.g. if I edit the hint text and then press "cancel".

Might be unrelated to this PR but "Make this question optional" doesn't seem to be plumbed in either, it always seems to be required.

model/src/form/form-editor/types.ts
```diff
@@ -277,16 +284,17 @@ export type FormEditorInputQuestionDetails = Pick<
...
+  questionDetails?: Partial<ComponentDef>
+  editRow?: {
     radioId?: string
     radioLabel?: string
     radioHint?: string
     radioValue?: string
     expanded?: boolean
   }
-  listItems: { label?: string; hint?: string; value?: string; id?: string }[]
+  listItems?: { label?: string; hint?: string; value?: string; id?: string }[]
```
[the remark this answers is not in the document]
alexluckett: actually, ignore the `page` part. I think that might not work with the future multiple-components per page model.

designer/server/src/common/components/editor-card/template.njk
```diff
@@ -22,8 +22,10 @@
                     }) }}
                   {% endif %}
 
-                  {% if params.caption %}
+                  {% if not params.noTitlePadding %}
```
alexluckett, line 25: looks like this is the only time `noTitlePadding` is used. Do we need the condition?

jbarnsley10: It's called only once (from radio-or-checkobx-edit.njk) to prevent an extra bit of space at the top of the container

designer/server/src/routes/forms/editor-v2/question-details.js
```diff
@@ -60,25 +109,39 @@ export default [
       const { yar } = request
       const { params, auth, query } = request
       const { token } = auth.credentials
-      const { slug, pageId, questionId } =
-        /** @type {{ slug: string, pageId: string, questionId: string }} */ (
-          params
-        )
```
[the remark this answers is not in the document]
jbarnsley10: Yep, had done that already on a later revision

designer/server/src/common/components/editor-card/template.njk
jbarnsley10: Have achieved this in a different way without needing the new param

designer/server/src/common/constants/session-names.js
```diff
@@ -4,7 +4,7 @@ export const sessionNames = {
   fileDownloadPassword: /** @type {const} */ ('fileDownloadPassword'),
   questionType: /** @type {const} */ ('questionType'),
   reorderPages: /** @type {const} */ ('reorderPages'),
-  enhancedActionState: /** @type {const} */ ('enhancedActionState'),
+  questionSessionState: /** @type {const} */ ('questionSessionState'),
```
whitewaterdesign, line 7: nice, think this is better. 👍

designer/server/src/lib/editor.js
```diff
@@ -47,16 +56,17 @@ export async function addPageAndFirstQuestion(
   questionDetails,
   pageDetails
 ) {
-  const requestUrl = new URL(
-    `./${formId}/definition/draft/pages`,
-    formsEndpoint
-  )
+  questionDetails.name = questionDetails.name ?? randomId()
```
whitewaterdesign, line 59: not a fan 😆

designer/server/src/lib/editor.js
```diff
@@ -47,16 +56,17 @@ export async function addPageAndFirstQuestion(
...
-  const { body } = await postJsonByType(requestUrl, {
+  const { body } = await postJsonByType(buildRequestUrl(formId, 'pages'), {
     payload: {
       title: pageDetails?.title ?? '',
       path: `/${slugify(pageDetails?.title ?? questionDetails.title)}`,
-      components: [questionDetails],
+      components: [
+        {
+          ...questionDetails
+        }
```
whitewaterdesign, line 68: you could just put that line here ^

jbarnsley10: Changed

designer/server/src/lib/list.js
```diff
@@ -15,31 +16,27 @@ const putJsonByListType =
...
-    title: payload.question,
+    name: stringHasValue(questionDetails.list)
+      ? questionDetails.list
+      : randomId(),
+    title: `List for question ${questionDetails.name}`,
     type: 'string',
-    items: payload.autoCompleteOptions
+    items: listItems.map((item) => {
+      return {
+        // id: item.id,
```
whitewaterdesign, line 32: model now creates a random uuid if none exist - so this can either be removed or uncommented.

designer/server/src/lib/redirect-helper.js
```diff
@@ -7,10 +7,18 @@ import { addErrorsToSession } from '~/src/lib/error-helper.js'
...
  * @param {Error | undefined} error
  * @param {ValidationSessionKey} errorKey
+ * @param {boolean} [removeAnchor]
  */
-export function redirectWithErrors(request, h, error, errorKey) {
+export function redirectWithErrors(request, h, error, errorKey, removeAnchor) {
   addErrorsToSession(request, error, errorKey)
   const { pathname: redirectTo } = request.url
+  if (removeAnchor) {
+    const [withoutAnchor] = redirectTo.split('#')
```
whitewaterdesign, line 16: like this 👍

designer/server/src/lib/redirect-helper.js
```diff
@@ -7,10 +7,18 @@ import { addErrorsToSession } from '~/src/lib/error-helper.js'
...
+    return h
+      .redirect(`${withoutAnchor}#`)
```
whitewaterdesign, line 18: fine to go for this PR, but would be worth asking @davidjamesstone if there's a better way to handle this in hapi.

designer/server/src/lib/list.js
jbarnsley10: Changed, having re-synced models

designer/server/src/models/forms/editor-v2/question-details.js
```diff
@@ -225,23 +236,23 @@ export function questionDetailsViewModel(
     getExtraFields(questionFieldsOverride, validation)
   )
 
-  validation = overrideFormValuesForEnhancedAction(
-    validation,
-    enhancedActionState
-  )
+  validation = overrideFormValuesForEnhancedAction(validation, state)
```
whitewaterdesign, line 239: agree this is a valid side effect. 👍

whitewaterdesign (approved):
Looks good - suggest the session logic could be encapsulated in a class to make it more reusable...but that's one for a refactor.  Good stuff, well done for all the hard work on this.

designer/server/src/lib/redirect-helper.js
whitewaterdesign: to follow on from this, we had a call, @davidjamesstone suggests removing line 19 `.code(StatusCodes.SEE_OTHER)` as it will use a different redirect that may remove the anchor.  Also that line 16 is unlikely to work anyway as there would never be a anchor on the pathname.  So this could be removed.

PR #790, opened 2025-04-02
Feat/481476  autocomplete question

whitewaterdesign:
## Autocomplete functionality

Non-JS version of autocomplete.  Apologise for the size, thinnest slice could do for this.  Appreciate could have broken it up.  Still to do - some validation errors not showing.  Aim to fix this in next PR.

Appreciate the parsing could have been separated out into another PR, but this PR does the following:

- creates a new joi extension `.dsv()` parser that takes a string and outputs an array - it allows delimiter separated values with custom line break and delimiter

### DIFF
--- a/model/src/form/form-definition/index.ts
+++ b/model/src/form/form-definition/index.ts
@@ -62,11 +62,19 @@ const conditionFieldSchema = Joi.object<ConditionFieldData>()
       .description('Human-readable name of the field for display purposes')
   })
 
-const conditionValueSchema = Joi.object<ConditionValueData>().keys({
-  type: Joi.string().required(),
-  value: Joi.string().required(),
-  display: Joi.string().required()
-})
+const conditionValueSchema = Joi.object<ConditionValueData>()
+  .description('Value specification for a condition')
+  .keys({
+    type: Joi.string()
+      .required()
+      .description('Data type of the value (e.g., string, number, date)'),
+    value: Joi.string()
+      .required()
+      .description('The actual value to compare against'),
+    display: Joi.string()
+      .required()
+      .description('Human-readable version of the value for display purposes')
+  })
 
 const relativeDateValueSchema = Joi.object<RelativeDateValueData>()
   .description('Relative date specification for date-based conditions')


designer/server/src/__stubs__/form-definition.js
```diff
@@ -4,6 +4,151 @@ import {
...
+      },
+      {
+        text: 'Java',
+        value: 'java'
+      }
+    ],
+    ...partialList
+  }
+}
+
```
whitewaterdesign, line 151: mainly moving this section up

designer/server/src/__stubs__/form-definition.js
```diff
@@ -4,6 +4,151 @@ import {
...
+    ...definitionPartial
+  }
+}
+
+/**
+ *
+ * @param {Partial<Item>} partialListItem
+ * @returns {Item}
+ */
+export function buildListItem(partialListItem = {}) {
```
whitewaterdesign, line 105: this is new

designer/server/src/lib/editor.test.js
```diff
@@ -20,36 +28,9 @@ import {
   setPageHeadingAndGuidance,
   updateQuestion
 } from '~/src/lib/editor.js'
-import { delJson, patchJson, postJson, putJson } from '~/src/lib/fetch.js'
 
 jest.mock('~/src/lib/fetch.js')
 
-const mockedDelJson = /** @type {jest.MockedFunction<typeof delJson>} */ (
```
whitewaterdesign, line 27: pulled this into a shared __stubs__ file

designer/server/src/lib/utils.test.js
```diff
@@ -1,31 +1,113 @@
...
+  mapListToAutoCompleteStr
+} from '~/src/lib/utils.js'
 
 jest.mock('@defra/hapi-tracing')
 
-describe('Header helper functions', () => {
-  it('should include the trace id in the headers if available', () => {
-    jest.mocked(getTraceId).mockReturnValue('my-trace-id')
+describe('utils', () => {
+  describe('Header helper functions', () => {
```
whitewaterdesign, line 21: this is not new, just wrapped in a descr block

designer/server/src/models/forms/editor-v2/advanced-settings-fields.js
```diff
@@ -399,11 +400,29 @@ export function getAdditionalSchema(payload) {
 
 /**
  * @param {Partial<FormEditorInputQuestion>} payload
+ * @returns {Partial<FileUploadFieldComponent>}
  */
-export function mapQuestionDetails(payload) {
+export function mapFileUploadQuestionDetails(payload) {
```
whitewaterdesign, line 405: from here down is a refactor to make it easier to use the list question details mapper

whitewaterdesign:
Note on coverage - most of that is code from previous cases that were never covered. Probably need to add tests for this at some point.

designer/server/src/lib/list.js
```diff
@@ -0,0 +1,127 @@
...
+    `./${formId}/definition/draft/lists/${listId}`,
+    formsEndpoint
+  )
+  const { body } = await putJsonByListType(addListUrl, {
+    payload: list,
+    ...getHeaders(token)
+  })
+
+  return body
+}
```
mokhld, line 86: @whitewaterdesign Should we add some error handling to these methods, or is it allowed to bublbe up and caught elsewhere?

designer/server/src/models/forms/editor-v2/advanced-settings-fields.js
mokhld: I'm wondering if we should rename the file itself to make it clearer where this is? Like advance-setting-field-mapper.js. This is a bit wordy though. In TS we would just do a `x.mapper.ts`

model/src/form/form-definition/index.ts
```diff
@@ -62,19 +62,11 @@ const conditionFieldSchema = Joi.object<ConditionFieldData>()
...
-      .description('The actual value to compare against'),
-    display: Joi.string()
-      .required()
-      .description('Human-readable version of the value for display purposes')
-  })
+const conditionValueSchema = Joi.object<ConditionValueData>().keys({
+  type: Joi.string().required(),
+  value: Joi.string().required(),
+  display: Joi.string().required()
+})
```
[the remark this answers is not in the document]
whitewaterdesign: oops, sorry, merge conflict resolution gone wrong.

designer/server/src/models/forms/editor-v2/advanced-settings-fields.js
whitewaterdesign: I think you're right, it probably even shouldn't be advanced-settings.  Will discuss this with @jbarnsley10 , we'll try and come up with a better name - maybe worth doing after his PR though to avoid conflicts.

designer/server/src/__stubs__/form-definition.js
alexluckett: I'm confused, what is this doing? Seems you're just returning what you've already passed in?

from the rest of this PR: 
```javascript
          buildListItem({
            text: 'Haskell',
            value: 'haskell'
          })
```

alexluckett: `buildList`, for example, doesn't use it

designer/server/src/lib/list.js
whitewaterdesign: so on research it is not handled in the hapi route handler.  We need to think about a pattern for this.  Thanks @mokhld , good spot.  Suggest leaving for future PR like other issue though.

designer/server/src/__stubs__/form-definition.js
whitewaterdesign: these are for building mocks.  It's a testing pattern where you have factories that build out stubs for you - so you only need to put in the fields that are appropriate for the specific test.  It means when you change the interface for that object you only have to change it in one place.  The alternative is trailing through a huge list of TS errors every time you make a change to an interface.

whitewaterdesign: https://dev.to/davelosert/mock-factory-pattern-in-typescript-44l9 here's a bit of description.

whitewaterdesign: @alexluckett ^

designer/server/src/lib/list.js
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Deletes a list
+ * @param {string} formId
+ * @param {string} listId
+ * @param {string} token
+ */
+export async function deleteList(formId, listId, token) {
```
alexluckett, line 94: Unused function. Should probably be plumbed into the delete handler?

e.g. when deleting a page, if the list is only referenced by this one question, delete the orphaned list too.

model/src/form/form-editor/index.ts
```diff
@@ -291,7 +291,153 @@ export const classesSchema = Joi.string()
...
+              Array.isArray(value) && value.every((k) => typeof k === 'string'),
+            message: 'must be an array of strings'
+          }
+        ]
+      }
+    }
+  }
+}) as CustomValidator
+
+export const autoCompleteOptionsSchema = customValidator
```
alexluckett, line 421: Love this. Great bit of engineering 🚀

designer/server/src/__stubs__/form-definition.js
alexluckett: Ah I see, thanks. Gotcha.

designer/server/src/lib/list.js
whitewaterdesign: good point...is this something that could be done as a subsequent PR?

alexluckett: Happy to. https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/535358 added to the board.

model/src/form/form-editor/types.ts
```diff
@@ -296,9 +300,14 @@ export interface GovukField {
   fieldset?: {
     legend?: { text?: string; isPageHeading?: boolean; classes?: string }
   }
-  value?: string | boolean | number | string[]
+  value?: string | boolean | number | string[] | Item[]
```
jbarnsley10, line 303: Is the addition of 'string[] and Item[]' correct? The field itself will still be a textarea, therefore 'string' ??

designer/server/src/lib/list.js
alexluckett: https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/535364 added

alexluckett (approved):

model/src/form/form-editor/types.ts
whitewaterdesign: once it's gone through validation it goes to `Item[]` so as far as the codebase is concerned this is the type - everything fails otherwise

PR #798, opened 2025-04-08
Fix/529912 duplicate page name

jbarnsley10:
Handles 409 errors from forms-manager

NOTE - TODO - rather than relying on part of the forms-manager error message being a particular string, in a new PR, I will implement something like a FUNCTION_CODE being returned in the error (in addition to the message text) so the designer can match against FUNCTION_CODEs. I have added ApiErrorFunctionCode enum in the model in preparation for this.

As part of this PR, I identified that the forms-manager is not throwing a 409 when the page title is UPDATED to a duplicate - only when a new page is being added that creates a duplicate page path. This will be tackled in the a new PR.

### DIFF
--- a/designer/server/src/lib/__stubs__/editor.js
+++ b/designer/server/src/lib/__stubs__/editor.js
@@ -1,3 +1,5 @@
+import Boom from '@hapi/boom'
+
 import config from '~/src/config.js'
 import { delJson, patchJson, postJson, putJson } from '~/src/lib/fetch.js'
 
@@ -27,6 +29,17 @@ export const formsEndpoint = new URL('/forms/', config.managerUrl)
 export const token = 'someToken'
 export const baseOptions = {
   headers: { Authorization: `Bearer ${token}` }
+}
+
+/**
+ * @param {string} message
+ */
+export function buildBoom409(message) {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  return Boom.boomify(new Error(message), {
+    statusCode: 409,
+    data: { message, statusCode: 409 }
+  })
 }
 
 /**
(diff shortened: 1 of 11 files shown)


designer/server/src/lib/error-boom-helper.js
```diff
@@ -0,0 +1,68 @@
...
+  const { error } = Joi.object({
+    [fieldName]: Joi.forbidden().messages({
+      '*': message
+    })
+  }).validate({ [fieldName]: 'force-error' })
+  return error
+}
+
+/**
+ * @param {Boom.Boom<{ message: string, statusCode: number }>} boomError
```
whitewaterdesign, line 38: Like this, think it's a really neat, well engineered solution.  Interested whether the Joi approach will work, might have a go in a bit, but if it doesn't think this is very solid.

designer/server/src/lib/__stubs__/editor.js
```diff
@@ -29,6 +31,17 @@ export const baseOptions = {
   headers: { Authorization: `Bearer ${token}` }
 }
 
+/**
+ * @param { string | undefined } message
+ */
+export function buildBoom409(message) {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  return Boom.boomify(new Error(message), {
```
whitewaterdesign, line 39: what was the issue here?  Upstream types?

designer/server/src/lib/error-boom-helper.js
```diff
@@ -0,0 +1,68 @@
...
+export function checkBoomError(boomError, errorKey, fieldName = 'general') {
+  if (!Boom.isBoom(boomError)) {
+    return undefined
+  }
+
+  const boomMessage = /** @type {string} */ (
+    boomError.data?.message ?? 'An error occurred'
+  )
+
+  const error = boomMappings.filter(
```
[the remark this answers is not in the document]
jbarnsley10: Changed to use a 'find'. Using your above caused linting errors due to it thinking 'error' was never undefined

whitewaterdesign (approved):
Great stuff @jbarnsley10 👍

PR #806, opened 2025-04-10
Redirect the user to the new URL if they're using the old one

alexluckett:
To support the new URL, redirect the users to it if they load the old URL.


alexluckett:
I've found another way that's easier. I'll abandon this.

PR #815, opened 2025-04-15
Feat/532357 preview errors

jbarnsley10:
Adds 'Preview error messages' buttons as appropriate
Changed hint text of short description to suggest it should start with an uppercase letter. If not, the error previews don't all have the correct case.


whitewaterdesign (approved):
Looks good. 👍

PR #817, opened 2025-04-16
528772 error messages

jbarnsley10:
Moved file type definitions into forms-model so forms-runner can share them
Corrected a series of error messages based on Laura's review of my temporary wording


whitewaterdesign (approved):
Looks good, nice easy one.

PR #821, opened 2025-04-22
Saves page path correctly

jbarnsley10:


designer/server/src/lib/editor.js
```diff
@@ -117,15 +118,32 @@ export async function updateQuestion(
...
   questionDetails
 ) {
-  // Determine if page controller should change
   const page = getPageFromDefinition(definition, pageId)
+
+  // Are we editing first question on page where there is no page title specified?
+  const questions = getComponentsOnPageFromDefinition(
+    definition,
+    pageId
+  ).filter((comp, idx) => !(comp.type === ComponentType.Markdown && idx === 0))
```
davidjamesstone, line 127: I wasn't sure about the check for `Markdown` here. Should it be checking for `isFormComponent`?

designer/server/src/lib/editor.js
```diff
@@ -117,15 +118,32 @@ export async function updateQuestion(
...
-    // Update page controller
+  if (
+    origControllerType !== newControllerType ||
+    isFirstQuestionAndNoPageTitle
+  ) {
+    // Update page controller and/or page path
     await patchJsonByType(buildRequestUrl(formId, `pages/${pageId}`), {
       payload: {
-        controller: newControllerType ?? null
+        controller: newControllerType ?? null,
```
davidjamesstone, line 145: Can we use `undefined` or doesn't this need to be `null`?

designer/server/src/lib/editor.js
jbarnsley10: This is a bit of a pain but the new editor allows the user to set 'guidance' for the page (independent to any questions). So we store a markdown component as the first component in the page. It's a bit nasty TBH.
So if I'm wanting to know if the currently editing question is the first on the page, I have to exclude any markdown at position zero. Unfortunately Markdown is still a form component, I think.
It might be more sensible for me to put this into a method called say getNonGuidanceComponents() - or something as it's used a few times in the codebase.

designer/server/src/lib/editor.js
jbarnsley10: It has to be null here as the API recognises 'null' as meaning 'remove the controller attribute from the JSON, don't just set it to null'. It's special case that was devised a while ago.

jbarnsley10: Extracted method out into forms-model

designer/server/src/lib/editor.js
jbarnsley10: Extracted method out into forms-model

davidjamesstone: I think `Markdown` should be a `ContentComponent`, not a `FormComponent`.
Have/Are we implementing other `ContentComponents` e.g. `Details` in the new editor?

jbarnsley10: Hmmm... good question. I would assume so but don't know at this point in time as not seen any screens for that.

PR #826, opened 2025-04-29
Ensure page title when there is more than 1 question

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/474316

### DIFF
--- a/designer/server/src/routes/forms/editor-v2/questions.js
+++ b/designer/server/src/routes/forms/editor-v2/questions.js
@@ -3,6 +3,7 @@ import {
   pageHeadingAndGuidanceSchema,
   pageHeadingSchema
 } from '@defra/forms-model'
+import Boom from '@hapi/boom'
 import { StatusCodes } from 'http-status-codes'
 import Joi from 'joi'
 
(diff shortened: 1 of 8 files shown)


designer/server/src/routes/forms/editor-v2/helpers.js
```diff
@@ -0,0 +1,16 @@
+import * as forms from '~/src/lib/forms.js'
+import { getPageFromDefinition } from '~/src/lib/utils.js'
+
+/**
+ * Returns the metadata, definition and page for a form
+ * @param {string} slug - the form slug
+ * @param {string} pageId - the page id
+ * @param {string} token - the access token
+ */
+export async function getFormPage(slug, pageId, token) {
```
whitewaterdesign, line 10: I don't think we ever thought about adding something like this.  Great idea. 👍

designer/server/src/routes/forms/editor-v2/question-details.js
```diff
@@ -210,16 +213,28 @@ export default [
       const { token } = auth.credentials
       const { slug, pageId, questionId, stateId } = params
 
+      // Form metadata and page components
```
whitewaterdesign, line 216: nice refactor 👍

whitewaterdesign (approved):
Nice one Dave, looks good to me. 👍

designer/server/src/routes/forms/editor-v2/questions.js
```diff
@@ -98,8 +107,28 @@ export default [
       // Form metadata and page components
       const metadata = await forms.get(slug, token)
       const definition = await forms.getDraftFormDefinition(metadata.id, token)
+      const page = getPageFromDefinition(definition, pageId)
```
jbarnsley10, line 110: One suggestion is to move this code out into a single method that checks if the page exists and determines if a redirect is necessary. Then the POST handler flow just has two lines such as:
const redirectForPageError = checkIfPageTitleMissing(defintion, pageId, payload)
if (redirectForPagError) {
  return redirectWithErros( ...)
}

It keeps the handler cleaner but not a biggy

PR #828, opened 2025-04-30
Re-order items with screenreader

jbarnsley10:
Branched from main
Used the same 'autofocus' mechanism as for page re-ordering in order to get the screenreader to immediately read the button label
However, although it works, the screenreading didn't seem 100% reliable. Let's see how the testing goes.

### DIFF
--- a/designer/server/src/common/components/editor-card/_editor-card.scss
+++ b/designer/server/src/common/components/editor-card/_editor-card.scss
@@ -28,3 +28,121 @@ $defra-brand-colour: govuk-organisation-colour(department-for-environment-food-r
 .govuk-summary-card__content .govuk-summary-list__show_final_divider .govuk-summary-list__row:last-of-type {
   border-bottom: 1px solid $govuk-border-colour;
 }
+
+// Styles for the reorderable list
+.gem-c-reorderable-list {
+  list-style-type: none;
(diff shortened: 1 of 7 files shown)


designer/server/src/common/components/editor-card/_editor-card.scss
```diff
@@ -28,3 +28,121 @@ $defra-brand-colour: govuk-organisation-colour(department-for-environment-food-r
 .govuk-summary-card__content .govuk-summary-list__show_final_divider .govuk-summary-list__row:last-of-type {
   border-bottom: 1px solid $govuk-border-colour;
 }
+
+// Styles for the reorderable list
+.gem-c-reorderable-list {
```
davidjamesstone, line 33: These styles look to have been copied from a ruby gem. Is there value in keeping the `gem-c-` prefix, rather than adopting them into proper application style with an `app-` prefix?

whitewaterdesign: good point, probably makes sense

jbarnsley10: Pushed a change for this.

mokhld (approved):
🚀

PR #829, opened 2025-04-30
Use flashed formValues on page error

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/548056


alexluckett (approved):

jbarnsley10:
We just need to check that the functionality hasn't been impacted. I seem to remember there being a reason why I wrote a stringHasValue() since I think we want '' (empty string) to mean it falls back to the alternative
i.e. functionality should be as if '||' as opposed to '??'

PR #840, opened 2025-05-07
Feat/481094 reorder list js

jbarnsley10:
Drag and drop reorder of list items

Includes a call to the  backend to update state after reordering
Shows error page if saving to backend state fails


designer/client/src/javascripts/preview/list.js
```diff
@@ -0,0 +1,529 @@
...
+
+  get afterInput() {
+    if (!this.list.length) {
+      return {
+        afterInputs: {
+          html:
+            '<div class="govuk-inset-text"><p class="govuk-body">No items added yet.</p><div class="govuk-inset-text">' +
+            '<p class="govuk-body">No items added yet.</p>' +
+            '</div></div>'
+        }
```
whitewaterdesign, line 371: forgot about this, should come from ListQuestionElements.afterInputsHTML

designer/client/src/javascripts/preview/helper.js
```diff
@@ -0,0 +1,14 @@
...
+ * re-adding the stateId if required
+ * @param {string} href
+ * @param {string} path
+ * @param {boolean} [includeStateId]
+ */
+export function addPathToEditorBaseUrl(href, path, includeStateId = false) {
+  const urlBaseEnd = href.indexOf('/editor-v2/')
+  const urlLastSlash = href.lastIndexOf('/')
+  return includeStateId
+    ? `${href.substring(0, urlBaseEnd + 10)}${path}${href.substring(urlLastSlash + 1)}`
```
whitewaterdesign, line 12: can you do this with js URL? wondering if it would be slightly cleaner.

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+    )
+  }
+
+  /**
+   * @param {HTMLElement[]} elements
+   * @param {boolean} inReorderMode
+   */
+  setStyleOnChildren(elements, inReorderMode) {
+    elements.forEach((item) => {
+      const cursorStyle = inReorderMode ? 'move' : 'default'
```
whitewaterdesign, line 95: wondering if we can do this with class rather than style?  Maybe it cut down on code a bit...maybe not though. If it doesn't then don't worry.

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+  List,
+  ListEventListeners,
+  ListQuestionElements
+} from '~/src/javascripts/preview/list'
+
+const REORDER_BUTTON_HIDDEN = 'reorder-button-hidden'
+
+const OK_200 = 200
+
+export class ListSortableQuestionElements extends ListQuestionElements {
```
whitewaterdesign, line 14: not one for now, but if this gets big think we can separate out the elements from the update to the elements - we should probably wait for SonarCloud to complain though.

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+      return
+    }
+    source.focus()
+  }
+
+  /**
+   * @param {HTMLElement[]} elements
+   */
+  setButtonsForReorderMode(elements) {
+    this.addItemButton.style = 'display: none'
```
whitewaterdesign, line 151: also probably could be done with a class.

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      ListQuestionElements.getListElementValues
+    )
+    this._list = this.createListFromElements(newList)
+    this.render()
+    return this._list
+  }
+
+  updateStateInSession() {
+    const url = addPathToEditorBaseUrl(window.location.href, '/state/', true)
```
whitewaterdesign, line 314: yeah, we could definitely use new URL for this.

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+        return 'error'
+      })
+  }
+
+  static setupPreview() {
+    const elements = new ListSortableQuestionElements()
+    const radio = new ListSortable(elements)
+    radio.render()
+
+    return radio
```
whitewaterdesign, line 348: Listeners will be decoupled from Question in next iteration

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+
+  updateStateInSession() {
+    const url = addPathToEditorBaseUrl(window.location.href, '/state/', true)
+
+    const listElements = Array.from(this._list).map(([_name, value]) => ({
+      id: value.id,
+      text: value.text,
+      hint: value.hint?.text ? { text: value.hint.text } : undefined,
+      value: value.value
+    }))
```
whitewaterdesign, line 321: this could sit in another getter on the class as is data related

designer/client/src/javascripts/preview/list-sortable.js
```diff
@@ -0,0 +1,356 @@
...
+        if (resp.status !== OK_200) {
+          this._listElements.redirectToErrorPage()
+          return 'error'
+        }
+        return 'ok'
+      })
+      .catch((_err) => {
+        this._listElements.redirectToErrorPage()
+        return 'error'
+      })
```
whitewaterdesign, line 340: this tightly couples API operations with a class that behaves more like a model (albeit coupled with nunjucks).  Think this sits much better within the listener class (ideally would sit in separate class altogether).

whitewaterdesign (commented):
Reordering logic looks great.  Only two issues I see are with use of `display:block` on DOMElement.style and think making the Question class responsible for api calls is an anti-pattern.

designer/client/src/javascripts/preview.js
```diff
@@ -17,17 +21,11 @@ import { UkAddress } from '~/src/javascripts/preview/uk-address.js'
 
 export function showHideForJs() {
   const previewPanel = document.getElementById('preview-panel')
-  if (previewPanel) {
-    previewPanel.style = 'display: block'
-  }
+  showHtmlElement(previewPanel)
```
whitewaterdesign, line 24: very nice, love this ❤️

designer/client/src/javascripts/preview/helper.js
```diff
@@ -6,9 +9,28 @@
...
-  return includeStateId
-    ? `${href.substring(0, urlBaseEnd + 10)}${path}${href.substring(urlLastSlash + 1)}`
-    : `${href.substring(0, urlBaseEnd + 10)}${path}`
+  const [baseUrl, remainingPath] = href.split(EDITOR_V2_FOLDER)
+  const stateId =
+    includeStateId && remainingPath ? `${remainingPath.split('/').pop()}` : ''
+  return path.startsWith('/')
+    ? `${baseUrl}${EDITOR_V2_FOLDER_NO_TRAILING_SLASH}${path}${stateId}`
+    : `${baseUrl}${EDITOR_V2_FOLDER}${path}${stateId}`
+}
```
whitewaterdesign, line 18: nice, think this is more readable.

designer/client/src/javascripts/preview/list-sortable.js
alexluckett: Sonar-driven development™️

whitewaterdesign (changes requested):
Would be good to move api method to the listener.  Happy to do it myself after this is merged though.

designer/client/src/javascripts/preview/list-sortable.js
alexluckett: Could pass in a callback in the constructor to inject that behaviour in?

```
const callback = (listItems) => {
    return fetch(...........)
}

new ListSortable(onSaveCallback)
```

PR #847, opened 2025-05-08
Fix/preview failing build

whitewaterdesign:


Dockerfile
```diff
@@ -62,6 +64,10 @@ COPY --from=productionBuild --chown=node:node /home/node/app/model/dist ./model/
 COPY --from=productionBuild --chown=node:node /home/node/app/designer/package.json ./designer/package.json
 COPY --from=productionBuild --chown=node:node /home/node/app/designer/client/dist ./designer/client/dist
 COPY --from=productionBuild --chown=node:node /home/node/app/designer/server/dist ./designer/server/dist
+COPY --from=productionBuild --chown=node:node /home/node/app/designer/bin/precompile.js ./designer/bin/
+
+RUN touch ./designer/precompile-govuk-components.js
```
alexluckett, line 69: Why do we need this, if we're copying it in above?

jbarnsley10: Yep, puzzling. Does it look for a later version of the file - otherwise ignore it?

whitewaterdesign: it runs on postinstall, ideally we won't run it when `RUN npm ci --omit=dev` is called.

alexluckett: Sorry I'm still super confused. Why do need to create an empty file?

alexluckett (approved):

Dockerfile
alexluckett: Answered, it's so it doesn't fail when it runs this step a second time, we already do an `npm ci` above. This is a quick workaround to unblock the build, we'll find a nicer solution layer on.

PR #848, opened 2025-05-09
feat: listSchema validation messages for unique identifiers

mokhld:
## Ticket
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/549452

## Description
This PR introduces a small change to the `listSchema` by introducing a new error message

Screenshot 2025-05-09 at 10 24 31

### DIFF
--- a/model/src/form/form-definition/index.ts
+++ b/model/src/form/form-definition/index.ts
@@ -591,6 +591,10 @@ export const listSchema = Joi.object<List>()
         .items(stringListItemSchema)
         .unique('text')
         .unique('value')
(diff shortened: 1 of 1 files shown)


jbarnsley10 (commented):
Is this really a requirement? Would it not be quite feasible and legitimate that a user wants some duplicate values in a list e.g. United Kingdom = 'UK' as well as 'Great Britain' = 'UK' ?

jbarnsley10 (approved):
If just to change the error message, as you say, then obvs totally fine

PR #863, opened 2025-05-16
Add repeater text to the pages overview

davidjamesstone:
Add missing flow 5:

Flow 5 – Display Question Listing Page

GIVEN I can see the Question overview page

AND I have finished configuring all the question pages I need

WHEN I select the back to all pages button

THEN the question listing page is displayed as per link whereby there is a row at the bottom of the summary list of said page called:

**People can answer More than once**

https://www.figma.com/design/1A9Tthy0CJeqp2rGNSKRjH/Defra-Form-Builder-flows?node-id=13384-96824&t=SkztKcLLhVb3YlXQ-4

image

### DIFF
--- a/designer/server/src/models/forms/editor-v2/pages.js
+++ b/designer/server/src/models/forms/editor-v2/pages.js
@@ -79,11 +79,20 @@ export function mapQuestionRows(page) {
 
   const isSummary = page.controller === ControllerType.Summary
 
-  return components.map((comp, idx) =>
+  const rows = components.map((comp, idx) =>
     comp.type === ComponentType.Markdown
       ? mapMarkdown(comp, isSummary)
       : mapQuestion(comp, idx)
   )
+
+  if (page.controller === ControllerType.Repeat) {
+    rows.push({
+      key: { text: 'People can answer' },
+      value: { text: 'More than once' }
+    })
+  }
+
+  return rows
 }
 
 /**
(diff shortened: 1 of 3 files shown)


designer/server/src/models/forms/editor-v2/pages.js
```diff
@@ -79,11 +79,20 @@ export function mapQuestionRows(page) {
...
       ? mapMarkdown(comp, isSummary)
       : mapQuestion(comp, idx)
   )
+
+  if (page.controller === ControllerType.Repeat) {
+    rows.push({
+      key: { text: 'People can answer' },
+      value: { text: 'More than once' }
+    })
+  }
```
mokhld, line 93: Do these need to be in a var or enum etc @davidjamesstone or is it a case that they're used just here?

alexluckett (approved):

designer/server/src/models/forms/editor-v2/pages.js
davidjamesstone: They're currently only used here.

I agree though - we have quite a lot of constant strings in the new editor - we should probably create a ticket to go through and i18n them like we do for [the old designer ](https://github.com/DEFRA/forms-designer/tree/main/designer/client/src/i18n)

mokhld: Yeah that makes sense 👍

PR #868, opened 2025-05-19
Handle non-interactive autocomplete preview

alexluckett:


whitewaterdesign:
almost the same, here's the original plan:
https://github.com/DEFRA/forms-designer/pull/869

PR #883, opened 2025-05-29
Feat/559206  restrict repeat upload

whitewaterdesign:
## Restricts ability to select Supporting evidence (upload) radio and Page repeaters

Uses two model functions to check whether supporting evidence component can be added and whether a page can be set to be repeater page. 

#### Prevent supporting evidence component from being added:

- **if page is a repeater page 🆕**
- if there are already more than one components on the page
- if there is already a file upload component on the page
- if the page is a new page and there's already one component (this rule is not in `omitFileUploadComponent` as is a purely designer concern - prior to the page being created)

#### Prevent page from being set to repeater page

- **if page is a file upload page 🆕**
- **if a file upload field appears on the page (defensive - I doubt this rule will ever be reached) 🆕**

This completes first and second points from this ticket:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/559206

Next step is the schema.


model/src/pages/helpers.ts
```diff
@@ -113,3 +114,45 @@ export function controllerNameFromPath(nameOrPath?: ControllerType | string) {
   const options = ControllerTypes.find(({ path }) => path === nameOrPath)
   return options?.name
 }
+
+export function canSetRepeater(
+  page: Page
+): page is Exclude<PageFileUpload, Page> {
+  if (page.controller === ControllerType.FileUpload) {
```
davidjamesstone, line 121: You should only allow to set to Repeater if the `controller === ControllerType.Page`

whitewaterdesign: Thanks @davidjamesstone , didn't know that.  Very helpful

alexluckett (approved):

model/src/pages/helpers.ts
```diff
@@ -113,3 +114,45 @@ export function controllerNameFromPath(nameOrPath?: ControllerType | string) {
...
+  if (!hasComponents(page)) {
+    return false
+  }
+  if (page.components.length > 1) {
+    return true
+  }
+  if (
+    page.components.some(
+      (component) => component.type === ComponentType.FileUploadField
+    )
```
mokhld, line 153: We're doing the same thing on line 126-128 above, shall we make a small helper for this to keep it DRY?

PR #885, opened 2025-05-29
Feat/559206  restrict repeat upload schema

whitewaterdesign:
## Restricts ability to select Supporting evidence (upload) radio

Following rules now added to the schema:

### When page controller is `FileUploadPageController` (v2 schema only)

1. One component must be `FileUploadField`
2. Additional component can only be of type `ContentComponentsDef` 
3. Min 1 component
4. Max 2 components

We don't need a check for Repeat exists within a file controller as when you save an upload field it sets the controller to be FileUploadPageController.

Resolves final point from: https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/559206
Chained from: https://github.com/DEFRA/forms-designer/pull/883


model/src/form/form-definition/index.ts
```diff
@@ -549,6 +594,19 @@ const eventsSchema = Joi.object<Events>()
       .description('Event handler triggered when the page data is saved')
   })
 
+export const pageUploadComponentsSchema = Joi.array<
+  FileUploadFieldComponent | ContentComponentsDef
+>()
+  .items(
+    fileUploadComponentSchema.required(),
```
whitewaterdesign, line 601: double check that content can be before.

model/src/form/form-definition/index.ts
```diff
@@ -449,6 +454,46 @@ export const componentSchemaV2 = componentSchema
...
+    .valid(ComponentType.Details)
+    .valid(ComponentType.Html)
+    .valid(ComponentType.Markdown)
+    .valid(ComponentType.InsetText)
+    .valid(ComponentType.List)
+    .required()
+    .description('Content only component type (Details, Html, Markdown, etc.)')
+})
+
+export const repeaterComponentSchema = componentSchema.keys({
```
whitewaterdesign, line 477: remove

alexluckett (approved):

PR #886, opened 2025-05-30
fix 551320: capitalise first letter on error preview

mokhld:
A fix following https://github.com/DEFRA/forms-designer/pull/871 to capitalise the first letter on the error descriptions:

Screenshot 2025-05-30 at 10 03 48

### DIFF
--- a/model/src/form/form-manager/types.ts
+++ b/model/src/form/form-manager/types.ts
@@ -5,6 +5,7 @@ export interface PatchPageFields {
   path?: string
   controller?: ControllerType | null
   repeat?: Repeat
+  condition?: string | null
 }
 
 export interface AddComponentQueryOptions {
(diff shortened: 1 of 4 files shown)


model/src/form/form-manager/types.ts
```diff
@@ -5,6 +5,7 @@ export interface PatchPageFields {
   path?: string
   controller?: ControllerType | null
   repeat?: Repeat
+  condition?: string | null
```
mokhld, line 8: This is a bit cheeky to include here, however I require it in the 'assign condition to a page' story.

davidjamesstone: It's a shame we're having to use null to tell forms-manager to unset the condition on the page. We need a better approach for this longer term

davidjamesstone (approved):

model/src/form/form-manager/types.ts
alexluckett: Ahhhhhhh

alexluckett: Thanks, makes sense

PR #903, opened 2025-06-06
Feature/556315 create simple condition

davidjamesstone:


whitewaterdesign:
Just sanity check due to challenges yesterday, does the migration handle these changes (v1 -> v2 and v2 -> v2)?

whitewaterdesign:
There are a lot of new components here that don't have any unit tests attached to them.  SonarCloud doesn't pick up on coverage from njk files so the coverage is probably a fair amount lower than the 91.5% described.  Especially given the complexity of what's being built reckon a bit more test coverage especially on these would be advisable.

designer/server/src/models/forms/editor-v2/condition.js
```diff
@@ -0,0 +1,331 @@
...
+      value="confirmSelectOperator">Select</button>`
+          }
+        },
+        ...insertValidationErrors(
+          validation?.formErrors[`items[${idx}].operator`]
+        )
+      }
+    : undefined
+
+  // TODO - enhance to handle date absolute + relative
```
whitewaterdesign, line 170: just checking that these are still TODO

designer/server/src/models/forms/editor-v2/condition.test.js
```diff
@@ -0,0 +1,167 @@
...
+          { text: 'Blue', value: '689d3f66-88f7-4dc0-b199-841b72393c19' },
+          { text: 'Green', value: '93d8b63b-4eef-4c3e-84a7-5b7edb7f9171' }
+        ],
+        name: 'items[2][value][itemId]',
+        value: '689d3f66-88f7-4dc0-b199-841b72393c19'
+      })
+    })
+
+    test('should return string value field', () => {
+      const stringItem = /** @type {ConditionDataV2} */ ({
```
whitewaterdesign, line 51: Suggestion: If these stubs are being used a lot might be worth adding a stub builder.

designer/server/src/models/forms/editor-v2/pages-helper.js
```diff
@@ -83,6 +88,26 @@ export function getFocus(focusStr) {
...
+  return {
+    page,
+    number: index + 1
+  }
+}
+
+/**
+ * @param {Page} page
+ */
+export function withConditionSupport(page) {
```
whitewaterdesign, line 105: might be worth making this into a type guard (and maybe even moving to model) like this:
TS
```
function isFish(pet: Fish | Bird): pet is Fish {
  return "swim" in pet;
}
```
jsdoc
```
/**
 * @param {Fish | Bird} pet
 * @returns {pet is Fish}
 */
function isFish(pet) {
  return "swim" in pet;
}
```

designer/server/src/routes/forms/editor-v2/condition.js
```diff
@@ -0,0 +1,330 @@
+import { randomUUID } from 'crypto'
```
whitewaterdesign, line 1: do wonder if it would make sense to move some of these helper methods out into another file, would make the route template a bit leaner.

I appreciate we haven't done it for other routes, but would make it easier when we come to refactor.

sonar-project.properties
```diff
@@ -11,7 +11,7 @@ sonar.javascript.lcov.reportPaths=coverage/lcov.info
 sonar.sourceEncoding=UTF-8
 
 sonar.sources=designer/client/src,designer/server/src,model/src,model/scripts
-sonar.exclusions=**/*.test.*,**/__stubs__/**,**/__mocks__/**,designer/client/src/javascripts/preview/nunjucks.js,designer/client/src/javascripts/application.js,designer/server/src/common/nunjucks/client.js
+sonar.exclusions=**/*.test.*,**/__stubs__/**,**/__mocks__/**,designer/client/src/javascripts/preview/nunjucks.js,designer/client/src/javascripts/application.js,designer/server/src/common/nunjucks/client.js,designer/server/src/models/forms/editor-v2/condition-value.js
```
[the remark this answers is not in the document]
jbarnsley10: File is excluded temporarily as it's not used yet. It's to be used as part of the date condition which is a later story

designer/server/src/routes/forms/editor-v2/condition.js
```diff
@@ -0,0 +1,330 @@
...
+
+const idSchema = Joi.string().uuid().allow('new').required()
+const stateIdSchema = Joi.string().optional()
+const componentIdSchema = conditionDataSchemaV2.extract('componentId')
+const operatorSchema = conditionDataSchemaV2.extract('operator')
+const valueSchema = conditionDataSchemaV2.extract('value')
+
+// Custom condition wrapper payload schema that
+// only allows conditions, not condition references
+const conditionWrapperSchema = conditionWrapperSchemaV2.keys({
```
mokhld, line 45: Is there any way to simplify this? Granted the nature of conditions is confusing, but this is quite hard to read and possibly making changes to?

davidjamesstone: Thanks @mokhld - I have added some more comments about how this schema works in [this commit](https://github.com/DEFRA/forms-designer/pull/903/commits/9080a1da19286297969c4cd69a36b2133b34cd92).

I'm don't think it can be made any simpler, but do [these extra comments](https://github.com/DEFRA/forms-designer/pull/903/commits/9080a1da19286297969c4cd69a36b2133b34cd92#diff-fe1abaecd9d9e7b5c3ae7b6c5cf7849139e5f283dc20936747a823e0639538e4R44) help?

alexluckett (approved):

PR #912, opened 2025-06-12
Feature 532794 & 532743: Upload and download json

mokhld:


designer/server/src/models/forms/editor-v2/upload.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+    pageCaption: {
+      text: pageCaption
+    },
+    errorList: buildErrorList(validation?.formErrors),
+    formErrors: validation?.formErrors,
+    formValues: validation?.formValues,
+    uploadAction: editorv2Path(metadata.slug, 'upload'),
+    downloadAction: `/library/${metadata.slug}/editor-v2/download`,
+    currentFormData: JSON.stringify(definition, null, 2),
```
alexluckett, line 42: Not sure what this is for - it's not used anywhere. Can be dropped I think?

designer/server/src/routes/forms/editor-v2/upload.js
```diff
@@ -0,0 +1,209 @@
...
+        const fileContent = await readStreamAsString(file)
+        definition = JSON.parse(fileContent)
+      } catch {
+        return flashErrorAndRedirect(
+          h,
+          yar,
+          slug,
+          ERROR_MESSAGES.INVALID_JSON_FILE
+        )
+      }
```
alexluckett, line 147: You can drop this if you set `route.options.payload.output: 'data'`. Hapi do the parsing for you.

https://hapi.dev/api/?v=21.4.0#route.options.payload.output

designer/server/src/routes/forms/editor-v2/upload.js
```diff
@@ -0,0 +1,209 @@
...
+
+      const fileName = file.hapi.filename
+      if (!fileName?.toLowerCase()?.endsWith('.json')) {
+        return flashErrorAndRedirect(
+          h,
+          yar,
+          slug,
+          ERROR_MESSAGES.INVALID_JSON_FILE
+        )
+      }
```
alexluckett, line 134: Rather than doing this yourself, you can set route.options.payload.allow to only allow JSON files. That way, by the time you receive `data` you'll have that guarantee it's a JSON file.

https://hapi.dev/api/?v=21.4.0#route.options.payload.allow

designer/server/src/views/forms/editor-v2/pages.njk
```diff
@@ -24,24 +25,56 @@
...
+                  {{ govukButton({
+                    text: action.text,
+                    href: action.href,
+                    classes: action.classes + " govuk-!-margin-bottom-0",
+                    attributes: action.attributes + " style='white-space: nowrap; padding: 12px 18px; font-size: 18px; min-width: auto; margin-bottom: 0;'",
+                    element: "a"
+                  }) }}
+                {% endfor %}
+              </div>
+            {% endif %}
```
alexluckett, line 48: Great stuff. Can we pull the inline styles out into `designer/client/src/stylesheets/_editor2.scss`?

designer/server/src/routes/forms/editor-v2/upload.js
mokhld: 🧠 
I wish i'd known this sooner. Makes the testing so much easier as well as the streams are handled by Hapi.

alexluckett: Probably only worth it if you can also look at `route.options.payload.allow` because I think you'll lose access to metadata like filename, because once you receive `data` it'll already have been parsed - it won't be a file anymore.

designer/server/src/routes/forms/editor-v2/upload.js
alexluckett: As discussed, hapi doesn't seem to make use of `failAction` when using the preferred solution so we must handle this in the handler to customise our error messages.

designer/server/src/routes/forms/editor-v2/upload.js
alexluckett: As discussed, hapi doesn't seem to make use of failAction when using the preferred solution so we must handle this in the handler to customise our error messages.

designer/server/src/models/forms/editor-v2/upload.js
alexluckett: GitHub highlighted 4 rows to give visual context but it's only `currentFormData` I was queryign specifically as I saw it was unused

PR #918, opened 2025-06-13
Feature/566740 number value conditions

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/566740

Used for:

NumberField, is
NumberField, is not
NumberField, is at least
NumberField, is at most
NumberField, is less than
NumberField, is more than

TextField, is longer than
TextField, is shorter than
TextField, has length

MultilineTextField, is longer than
MultilineTextField, is shorter than
MultilineTextField, has length

EmailAddressField, is longer than
EmailAddressField, is shorter than
EmailAddressField, has length

Also [fixes](https://github.com/DEFRA/forms-designer/pull/918/files#diff-21b23afe3b549db03de0178364940fef5abed27fe3b1ea822a840e4e73e936aaR99) YesNoField should use BooleanValue

### DIFF
--- a/designer/server/src/models/forms/editor-v2/condition-value.js
+++ b/designer/server/src/models/forms/editor-v2/condition-value.js
@@ -147,82 +147,25 @@ export function buildValueField(
 ) {
   switch (type) {
     case ConditionType.ListItemRef: {
-      return {
-        id: `items[${idx}].value`,
-        name: `items[${idx}][value][itemId]`,
-        fieldset: {
-          legend: {
-            text: 'Select a value'
-          }
-        },
-        classes: GOVUK_RADIOS_SMALL,
-        value:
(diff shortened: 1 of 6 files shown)


designer/server/src/models/forms/editor-v2/condition-value.js
```diff
@@ -216,6 +178,109 @@ export function buildValueField(
...
+    label: {
+      text: 'Enter a value'
+    },
+    classes: 'govuk-input--width-5',
+    attributes: {
+      inputmode: 'numeric'
+    },
+    value:
+      'value' in item && 'value' in item.value
+        ? item.value.value.toString()
```
jbarnsley10, line 278: Could item.value.value be undefined, thereby throwing an error on the toString()?

PR #927, opened 2025-06-17
feat: 552374 - Pull in PageController model

whitewaterdesign:
### Page controller model

This is the first part of the changes for the new Page Preview - just the model initially - the other changes are on another branch.  Uses a page controller which goes through the components and maps them to `Question`s.  A couple of new components have been added to the list as we need Markup for the guidance text.

Guidance now works through the component - so this needed to be represented in the page controller, also there's a dummy text the moment you highlight the guidance text - so we are creating two separate guidance components and switching between them and no guidance component.

This ticket
[https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/552374](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/552374)

### DIFF
--- a/model/src/form/form-editor/preview/autocomplete.js
+++ b/model/src/form/form-editor/preview/autocomplete.js
@@ -25,17 +25,7 @@ export class AutocompleteListQuestion extends Question {
   }
 }
 
-export class AutocompleteQuestion extends Question {
-  /**
-   * @type {ComponentType}
-   */
-  componentType = ComponentType.AutocompleteField
-  _questionTemplate = PreviewComponent.PATH + 'autocompletefield.njk'
-  /**
-   * @type {string}
-   * @protected
-   */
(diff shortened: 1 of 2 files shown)


model/src/form/form-editor/preview/autocomplete.js
```diff
@@ -1,8 +1,36 @@
+import { ComponentType } from '~/src/components/enums.js'
 import { autoCompleteOptionsSchema } from '~/src/form/form-editor/index.js'
+import { PreviewComponent } from '~/src/form/form-editor/preview/preview.js'
 import { Question } from '~/src/form/form-editor/preview/question.js'
 
+export class AutocompleteListQuestion extends Question {
```
mokhld, line 6: Quite a lot of overlap with `AutocompleteQuestion` below. Use composition or inheritance?

PR #937, opened 2025-06-20
Feature/571049 duplicate condition display name

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/571049

### DIFF
--- a/designer/server/src/routes/forms/editor-v2/condition.js
+++ b/designer/server/src/routes/forms/editor-v2/condition.js
@@ -2,6 +2,7 @@ import { randomUUID } from 'crypto'
 import Stream from 'node:stream'
 
 import {
+  FormDefinitionError,
   conditionDataSchemaV2,
   conditionWrapperSchemaV2,
   slugSchema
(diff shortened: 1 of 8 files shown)


designer/server/src/routes/forms/editor-v2/condition.js
```diff
@@ -183,19 +188,37 @@ export default [
...
+        ) {
+          const joiErr = createJoiError(
+            'displayName',
+            'Duplicate condition name'
+          )
+
+          return redirectWithErrors(request, h, joiErr, errorKey, '#')
+        }
+
+        throw err
```
alexluckett, line 220: do we want to add a generic error handler here as a fallback for all the other errors?

alexluckett: e.g.

```
# Your form could not be saved

{{ errors }}
```

alexluckett: that way if we forget to catch one in a specific route, we at least have a fallback

alexluckett: All good, happy to pick that up separately

alexluckett: https://dev.azure.com/defragovuk/DEFRA-CDP/_boards/board/t/DXT%20Engineering/Stories?workitem=572794 raised

PR #940, opened 2025-06-23
Use conditional alternatives for conditions

davidjamesstone:
Related to https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/571049

### DIFF
--- a/model/src/form/form-definition/index.ts
+++ b/model/src/form/form-definition/index.ts
@@ -368,9 +368,11 @@ export const conditionWrapperSchemaV2 = Joi.object<ConditionWrapperV2>()
       ),
     items: Joi.array<ConditionGroupDataV2>()
       .items(
-        Joi.alternatives()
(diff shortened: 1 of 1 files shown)


model/src/form/form-definition/index.ts
```diff
@@ -368,9 +368,11 @@ export const conditionWrapperSchemaV2 = Joi.object<ConditionWrapperV2>()
       ),
     items: Joi.array<ConditionGroupDataV2>()
       .items(
-        Joi.alternatives()
-          .try(conditionDataSchemaV2, conditionRefDataSchemaV2)
-          .required()
+        Joi.alternatives().conditional('.componentId', {
```
jbarnsley10, line 371: Do you need the leading dot? The examples I saw just referenced the key name

davidjamesstone: It's equivalent with or without the dot. Relative refs can start with a dot https://joi.dev/api/?v=17.13.3#Relative-references

PR #951, opened 2025-06-24
Feat/552374  page overview client preview

whitewaterdesign:
Final part (hopefully) of the client side render.

Pulls in all the scaffolding from model and sets up to run in JS.

image

Checked coverage - I can improve on it tomorrow morning.

**Tickets**:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/552374
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/551824


designer/client/nunjucks-loader.js
```diff
@@ -15,10 +15,15 @@ env.addFilter('markdown', markdown)
  */
 export default function (source) {
   // @ts-expect-error
-  const name = path.basename(this.resourcePath)
+  const resourcePath = /** @type {string} */ (this.resourcePath)
```
jbarnsley10, line 18: A comment to explain what is happening here would be helpful as not that easy to follow the pop's

designer/server/src/views/forms/editor-v2/partials/questions-right-panel.njk
```diff
@@ -5,3 +5,8 @@
 {% call appPagePreviewPanel(previewModel) %}
 {% endcall %}
 
+<script src="{{ getAssetPath("assets/nunjucks/govuk-components.js") }}"></script>
+<script type="module">
+    import { setupPageController } from "{{ getAssetPath("pagePreview.js") }}"
+    setupPageController({{ preview.page | safe }}, {{ preview.definition | safe }})
```
jbarnsley10, line 11: It seems a lot to pass the whole form definition into the JS as a param here, especially for a page preview

jbarnsley10 (approved):
Looks really good - nice work. Just a couple of comments

designer/server/src/views/forms/editor-v2/partials/questions-right-panel.njk
whitewaterdesign: Page preview includes all the components of the page, and it needs the lists too.  In future it needs sections, so although not ideal there's not really a good way around it.

jbarnsley10: Could we just send lists and sections, rather than re-adding all the pages info again? Just trying to reduce the footprint of loading this JS

PR #953, opened 2025-06-26
Fixed page title on check change condition

jbarnsley10:


model/src/pages/helpers.ts
```diff
@@ -153,3 +153,23 @@ export function omitFileUploadComponent(page: Page | undefined): boolean {
...
+  if (hasComponents(page) && page.components.length > 0) {
+    const firstComp = page.components.find(
+      (c) => c.type !== ComponentType.Markdown
+    )
+    if (firstComp) {
+      return firstComp.title
+    }
+  }
+  return 'Page title unknown'
+}
```
alexluckett, line 175:
```javascript
/**
 * Gets page title, or title of first question (if no page title set)
 * @param {Page} page
 * @returns {string}
 */
export function getPageTitle(page: Page) {
  if (page.title !== '') {
    return page.title
  }

  if (hasComponentsEvenIfNoNext(page)) {
    const inputComponents = page.components.find(hasFormField)

    if (inputComponents) {
      return inputComponents.title
    }
  }

  return 'Page title unknown'
}
```

some helpers

alexluckett (commented):
some helpers which might be useful

alexluckett (approved):

PR #970, opened 2025-07-04
Create conditions from the conditions tab

jbarnsley10:
Allows a user to view and create conditions from the conditions tab (previously this tab only gave the user a link to go to the conditions manager page)

- Refactored to re-use original conditions handlers and schemas


alexluckett (changes requested):
Sorry Jez, not quite right. Great refactor but this is not a 1:1 match with the advanced condition builder in the condition manager.

This condition is only meant to be max 1 depth, like this:

Image

The user wouldn't have the ability to add more than one condition.

alexluckett:
Just to be clear, all I'm saying is we need to hide this button on the "simple" route:

image

alexluckett (approved):

PR #979, opened 2025-07-09
Feat/df 67 repeater page preview

whitewaterdesign:
## Repeater on Page Preview

Adding repeater to page preview.  Also scaffolding in place for when we add section titles.

image

Ticket: https://eaflood.atlassian.net/browse/DF-67


designer/client/src/javascripts/preview/__stubs__/page.js
```diff
@@ -1,20 +1,186 @@
 export const pageHeadingAndGuidanceHTML = `
 <div>
-<div class="govuk-checkboxes__item">
-      <input class="govuk-checkboxes__input" id="pageHeadingAndGuidance" name="pageHeadingAndGuidance" type="checkbox" value="true" checked="">
+  <div class="govuk-checkboxes__item">
```
jbarnsley10, line 3: This is a lot of HTML to keep up-to-date with the NJK pages but I guess there's not an easy alternative

designer/server/src/common/components/preview-panel-layout/_preview-panel-layout.scss
```diff
@@ -68,6 +68,10 @@ $defra-brand-colour: govuk-organisation-colour(department-for-environment-food-r
   .highlight {
     border-bottom: 2px solid #ffb266;
   }
+
+  .govuk-button .highlight {
+    color: #000000;
```
jbarnsley10, line 73: Should this be govuk("black") ?

designer/client/src/javascripts/preview/__stubs__/page.js
whitewaterdesign: alternative would be to build out the DOM...but that would be a pain

PR #1003, opened 2025-07-18
Skip link goes to body content

jbarnsley10:
Body is wrapped in <div> so default skip link (going to #main-content) now goes to body content area


whitewaterdesign (approved):
LGTM

PR #1007, opened 2025-07-21
Feat/df 289 fetch definition

whitewaterdesign:
## Cleanup Page and Definition when setting up preview

For `setupPageController` and its siblings, when a page or definition is needed we are now doing an api call to collect these.

Links to: https://eaflood.atlassian.net/browse/DF-289


alexluckett (approved):
Great work, Chris! Thanks for the speed 🚀

PR #1014, opened 2025-07-23
Form create test snapshots

davidjamesstone:
Adds test snapshots to form create routes with a view to extending to other pages


mokhld:
> Interesting, I'd say lets try it and see how it goes, although could end up being brittle. When you make a change, how do you go about it?

If this is like jest snapshot testing, then assumedly you'd have to re-generate the snapshots with each UI change.

mokhld (approved):
🔥

PR #1020, opened 2025-07-29
Delete a user

jbarnsley10:
Delete a user

Ticket [DF-150](https://eaflood.atlassian.net/browse/DF-150)

### DIFF
--- a/designer/server/src/models/manage/users.js
+++ b/designer/server/src/models/manage/users.js
@@ -74,6 +74,9 @@ export function createOrEditUserViewModel(allRoles, user, validation) {
         ...insertValidationErrors(formErrors?.userRole)
       }
     },
+    displayName: {
+      text: user?.displayName
+    },
     userId: user?.userId
   }
 }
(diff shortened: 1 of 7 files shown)


designer/server/src/views/manage/user.njk
```diff
@@ -39,12 +40,7 @@
           text: 'Save changes'
         }) }}
 
-        <!-- to be implemented later
-        {{ govukButton({
-          text: 'Remove user',
-          classes: 'govuk-button--warning'
-        }) }}
-        -->
+        <a href="/manage/users/{{ userId }}/delete" id="remove-user-button" class="govuk-button govuk-button--warning">Remove user</a>
```
alexluckett, line 43: you can use the govukButton component I think, just pass in a `href` and it should convert it into an `<a>` element

designer/server/src/views/manage/user.njk
```diff
@@ -39,12 +40,7 @@
...
 
         <a href="/manage/users" class="govuk-link govuk-link--no-visited-state" id="cancel-edit-user">Cancel</a>
```
alexluckett, line 45: as above

designer/server/src/routes/manage/user.js
```diff
@@ -36,11 +42,47 @@ const userIdSchema = Joi.object({
...
   ({
     method: 'GET',
-    path: '/manage/users/{userId}/{amend?}',
+    path: `${MANAGE_USERS_BASE_URL}/new`,
+    async handler(request, h) {
+      const { auth, yar } = request
+      const { token } = auth.credentials
+
+      // Get list of possible roles
+      const roles = await getRoles(token)
```
alexluckett, line 58: out of scope of this bit of refactoring you did, but how are we mapping the role from the API to the description of what the role can do?

E.g.  "Form publisher" -> "Can create and publish new and existing forms"

jbarnsley10: Great point. I totally missed the hint text of the roles. It's defined in the entitlement API and retrieved via a call to getRoles()

designer/server/src/models/manage/users.js
```diff
@@ -68,12 +68,18 @@ export function createOrEditUserViewModel(allRoles, user, validation) {
         },
         items: allRoles.map((r) => ({
           text: r.name,
-          value: r.code
+          value: r.code,
+          hint: {
+            text: r.description
```
alexluckett, line 73: I'm thinking this we probably want this to be stored in the frontend instead, retrieved via a mapping - that way we keep the UI doing presentation with the backend doing just the data. If it ever came to building translations, this would allow us to support them. This is similar to what we're doing in manager, e.g. manager throws errors, UI prepares the actual error message.

PR #1025, opened 2025-07-31
DF-352: Fix guidance position during component reorder

davidjamesstone:
- Relax component name unique constraint by ignoring undefined. This will allow us to support multiple `ContentComponents` for plugin users.
- Make `name` field required for form components

### DIFF
--- a/designer/server/src/routes/forms/editor-v2/questions.js
+++ b/designer/server/src/routes/forms/editor-v2/questions.js
@@ -37,7 +37,10 @@ import {
 } from '~/src/models/forms/editor-v2/pages-helper.js'
 import * as viewModel from '~/src/models/forms/editor-v2/questions.js'
 import { editorv2Path } from '~/src/models/links.js'
-import { customItemOrder } from '~/src/routes/forms/editor-v2/helpers.js'
+import {
+  customItemOrder,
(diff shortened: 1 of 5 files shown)


designer/server/src/routes/forms/editor-v2/questions.js
```diff
@@ -207,6 +211,7 @@ export default [
         // Save re-order (if in reorder mode) in case user pressed the main 'Save changes' button
         // as opposed to the reorder 'Save changes' button
         if (action === 'reorder') {
+          mergeMissingComponentsIntoOrder(page, itemOrder)
```
alexluckett, line 214: should we move this to the top of the function, rather than calling it twice?

```javascript
const { movement, itemOrder, saveReorder } =
        /** @type {{ movement: string, itemOrder: string[], saveReorder: boolean}} */ (
          payload
)
mergeMissingComponentsIntoOrder(page, itemOrder)
```

alexluckett: well, after `const page` not right at the top, my bad

davidjamesstone: Note: We agreed to keep as is and refactor the handler in a future ticket.

PR #1026, opened 2025-08-01
Df 311 form edited event model

whitewaterdesign:


model/src/form/form-audit/index.ts
```diff
@@ -57,26 +58,38 @@ export const formCreatedMessageData =
...
+  FormDefinitionRequestType.UPDATE_PAGE_FIELDS,
+  FormDefinitionRequestType.DELETE_PAGE
+]
 
 export const formUpdatedMessageData =
-  formMessageDataBase.append<FormUpdatedMessageData>({
-    changeSet: formChangeSet.required()
+  formDefinitionMessageBase.append<FormUpdatedMessageData>({
+    payload: Joi.object().required(),
+    requestType: Joi.string().valid(...allowedDefinitionRequestTypes)
```
davidjamesstone, line 92: `.required()`?

model/src/form/form-audit/types.ts
```diff
@@ -130,10 +129,19 @@ export interface FormUploadedMessageData extends FormMessageDataBase {
   changes: ChangesMessageData<FormUploadedChanges>
 }
 
-export type FormChangeSet = IChange
+export interface FormDefinitionS3Meta {
+  fileId?: string
```
davidjamesstone, line 133: I think these 3 props should not be `?`. According to the schema they're `string().required()`

davidjamesstone (approved):

PR #1041, opened 2025-08-11
Types/enums for entitlement audit

jbarnsley10:
Types, enums and schemas for Entitlement audit events


model/src/form/form-audit/index.ts
```diff
@@ -292,8 +302,27 @@ export const messageSchema = Joi.object<AuditMessage>().keys({
       {
         is: Joi.string().trim().valid(AuditEventMessageType.FORM_UPDATED),
         then: formUpdatedMessageData
+      },
+      {
+        is: Joi.string()
+          .trim()
+          .valid(AuditEventMessageType.ENTITLEMENT_CREATED),
```
whitewaterdesign, line 309: up to you but you could put all three into the same check - if the `then` is the same

PR #1048, opened 2025-08-12
fix DF-401: Ensure user session is remove when deleting themselves

mokhld:


alexluckett (commented):
Isn't the wider problem that a users roles aren't refreshed upon save? I'm think we're tackling this at too low of a level.

I'd expect that when `addUser`, `updateUser`, `deleteUser` is called, we find the session and refresh the roles for any user, not just if they're updating themselves.

My suggestion:
- Pull these out into a userService.js with the three exported functions
- Add some private function(s) to handle the refresh of their permissions. These would go into redis and update the user session.
- Swap the route from calling lib directly, to calling the new service layer

Note: look in `designer/server/src/common/helpers/auth/` for some helpers we already to have to get/drop a user session. There's probably some small refactoring we can do to use the logic already there. E.g. they currently require a `request` which wouldn't be present in this scenario, so we might be able to split them out to `getUserSessionFromUserId`, `getUserSessionFromRequest`, etc.

For example:

```
export function updateUser(server, userId, roles) {
    await lib.updateUser(userId, roles)
    await refreshSessionPermissions(userId)
}

export function deleteUser(server, userId) {
    await lib.deleteUser(userId, roles)
    await refreshSessionPermissions(userId)
}

// designer/server/src/common/helpers/auth/refresh-user-session-entitlements.js
function refreshSessionPermissions(userId) {
    // look in redis, find the session
    // if user exists, refresh their role and scopes
    // if user does not exist in the remote API, drop it
}
```

the above scenario should handle any user update scenario, we don't need to handle individual users specifically

designer/server/src/common/helpers/auth/refresh-user-session-entitlements.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+/**
+ * @param {Server} server
+ * @param {string} userId
+ */
+export async function dropUserSessionById(server, userId) {
+  try {
+    await server.methods.session.drop(userId)
```
alexluckett, line 58: apologies, I forgot it was this simple - this helper is redundant because of the session helper on the server.

designer/server/src/routes/manage/user.js
```diff
@@ -209,26 +204,29 @@ export default [
...
-
         yar.flash(
           sessionNames.successNotification,
           `You updated ${existingUser.displayName}'s role to ${getNameForRole(userRole)}`
         )
 
-        // Redirect back to list of users
+        if (isSelfUpdate && isDowngradingToFormCreator) {
+          return h.redirect('/library').code(StatusCodes.SEE_OTHER)
+        }
```
alexluckett, line 228: Stil uneasy about this bit - feels too curated and brittle. What we add more roles, for example?

Something a bit more future-proof would be to check the scopes, given the latest update.

you could do something like:

**designer/server/src/common/helpers/auth/refresh-user-session-entitlements.js**
```javascript
refreshUserSessionEntitlements(request, userId, token) {
    // stuff
    const { scopes } = await getUser(token, userId)
    existingSession.scope = scopes

    if (request.credentials.userId === userId) {
        request.credentials.scope = scopes // won't be automaticallty updated by hapi until the next request, so patch the current request
    }
    // other stuff
}
```

**designer/server/src/routes/manage/user.js**
```javascript
try {
    await request.server.auth.verify(request) // this throws if the user is no longer entitlemed to access this endpoint
} catch (err) { 
    return h.redirect('/library').code(StatusCodes.SEE_OTHER)
}
```

with the above method, we no longer have to worry about if it's a self-update and what roles are impacted - hapi uses the existing RBAC system so we retain a single source of truth. Note I believe this relies on us patching the scopes on the request (since that was constructed before the role was updated).

designer/server/src/routes/manage/user.js
```diff
@@ -272,17 +270,12 @@ export default [
       try {
         const existingUser = await getUser(token, userId)
 
+        await userService.deleteUser(server, token, userId)
+
         if (isSelfDeletion) {
-          await server.methods.session.drop(userId)
           request.cookieAuth.clear()
         }
```
alexluckett, line 277: cleaner to move these lines into userService.deleteUser so we have one source of truth for the functionality

alexluckett (approved):

PR #1056, opened 2025-08-15
feat: download json audit event and tests

mokhld:
## Story link
https://eaflood.atlassian.net/browse/DF-305?atlOrigin=eyJpIjoiOTAzZmY1ODkwNThkNGEzMmI0NjRmYWZlNmEwZTFiYzYiLCJwIjoiaiJ9

### DIFF
--- a/designer/server/src/routes/forms/editor-v2/download.js
+++ b/designer/server/src/routes/forms/editor-v2/download.js
@@ -1,6 +1,8 @@
 import { Scopes } from '@defra/forms-model'
 
+import { mapUserForAudit } from '~/src/common/helpers/auth/user-helper.js'
 import * as forms from '~/src/lib/forms.js'
+import { publishFormDownloadedEvent } from '~/src/messaging/publish.js'
 
 export const ROUTE_PATH_DOWNLOAD = 'download'
 export const ROUTE_FULL_PATH_DOWNLOAD = '/library/{slug}/editor-v2/download'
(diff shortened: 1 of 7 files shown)


designer/server/src/routes/forms/editor-v2/download.js
```diff
@@ -24,6 +26,18 @@ export default [
       const filename = `${slug}.json`
       const definitionJson = JSON.stringify(definition, null, 2)
 
+      // Attempt to publish audit event, but don't fail the download if it fails
+      try {
+        const auditUser = mapUserForAudit(auth.credentials.user)
```
mokhld, line 31: I can't see us doing this with current audit publishing fields, but this will allow it to fail gracefully if it does for whatever reason.

whitewaterdesign: Do you want to fail gracefully?  
I have always considered auditing to be an atomic action - if auditing fails the transaction or action should fail because:

> ChatGPT: If you can commit them together, you avoid gaps (“state changed but no audit trail”) or ghosts (“audit shows change that never happened”).
> 
> Why atomic matters
> 
> Compliance & forensics: missing or out-of-order entries can make records inadmissible or untrustworthy.
> 
> Threat modeling: attackers often try to perform an action then block logging; atomicity defeats that.
> 
> Ops: debugging and reconciliation get much harder when audit and data can diverge.

We've taken that approach on manager, it is part of the transaction.

PR #1066, opened 2025-08-20
feat: Adaptor listener types

mokhld:


model/src/form/form-notify-listener/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  data: FormAdapterSubmissionMessageData
+}
+
+export interface FormAdapterSubmissionMessageMeta {
+  schemaVersion: FormAdapterSubmissionSchemaVersion
+  timestamp: Date
+  referenceNumber: string
+  formName: string
+  formId: string
+  status: FormStatus
```
whitewaterdesign, line 14: isPreview: boolean

PR #1094, opened 2025-09-17
bug DF-405: preserve repeater settings when updating question properties

mokhld:
https://eaflood.atlassian.net/browse/DF-405


mokhld:
> Thanks for adding the tests. My mistake that this code needed to cater for the uncheck of repeater.

No worries and thanks for reviewing!

PR #1096, opened 2025-09-19
feat DF-450:  identify conditions referencing specific component IDs and bet…

mokhld:
…ter error messaging

https://eaflood.atlassian.net/browse/DF-450?atlOrigin=eyJpIjoiOTg0ZWMzZWY3M2NjNGUwMjgzNGQ4MTdjZjQwYjY5NGMiLCJwIjoiaiJ9

### DIFF
--- a/designer/server/src/lib/deletion-helpers.js
+++ b/designer/server/src/lib/deletion-helpers.js
@@ -1,0 +1,101 @@
+import { isFormType } from '@defra/forms-model'
+
+import { formatList } from '~/src/common/helpers/format.js'
+import { findConditionsReferencingComponents } from '~/src/lib/condition-references.js'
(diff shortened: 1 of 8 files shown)


designer/server/src/lib/deletion-helpers.js
```diff
@@ -0,0 +1,181 @@
...
+ * @param {string} formId
+ * @param {string} token
+ * @param {string} pageId
+ * @param {FormDefinition} definition
+ */
+export async function performPageDeletion(formId, token, pageId, definition) {
+  await deletePage(formId, token, pageId, definition)
+}
+
+/**
```
jbarnsley10, line 152: I'm a little confused about this method. Surely if a questionId is supplied, you are deleting a question, otherwise you're deleting the page?

PR #1114, opened 2025-10-02
fix/DF-531: Error stack traces

davidjamesstone:
This should fix missing stack traces when an unhandled exception is encountered in a route handler.

[More info](https://discord.com/channels/1031830080109957162/1031830081242415106/threads/1423251797040042065)
---

Hi all 👋 

I'm looking for a way to capture and log stack traces when an unhandled exception is encountered in a route.

I have a generic onPreResponse handler for error logging that looks something like:

```
server.ext('onPreResponse', (request, h) => {
  const response = request.response

  if (Boom.isBoom(response)) {
    // Stack trace is undefined for unhandled system errors
    console.log(response.stack)
    
    return h.response({}).code(500)
  }
  
  return h.continue
})
```

If I have a route that generates an unhandled error e.g.

```
{
    method: 'GET',
    path: '/generate-error',
    async handler(request, h) {
      const o = undefined
      
      // Generate a TypeError
      const val = o.val
      
      return { some: 'thing' }
    }
}
```

The `response.stack` in `onPreResponse` is always undefined.

Stepping through the code it is because of [this line](https://github.com/hapijs/hapi/blob/master/lib/toolkit.js#L64) in toolkit.js uses Boom.badImplementation which is stripping out the stack [here](https://github.com/hapijs/boom/blob/master/lib/index.js#L74) in Hoek.clone

Is this by design? Has anyone overcome this and found a way to access the err.stack in an onPreResponse for unhandled system errors?

TIA

djs — 11:25 AM
Hoek is stripping it [here](https://github.com/hapijs/hoek/blob/master/lib/clone.js#L105) when cloning the error, it gets the descriptor with key "stack" and redefines the property on the cloned error Object.defineProperty(newObj, key, descriptor) but the descriptor value is lost

djs — 12:16 PM
Looking at the most recent version of Hoek (11.0.7) it looks like stack now gets [special](https://github.com/hapijs/hoek/blob/master/lib/clone.js#L179) [handling](https://github.com/hapijs/hoek/blob/master/lib/clone.js#L92) (in node 21+). If I force Boom to use Hoek version 11.0.7, I see a stack trace in onPreResponse.

Latest Boom [uses](https://github.com/hapijs/boom/blob/master/package.json#L21) Hoek@^11.0.2.

Ok, I've now just seen [this](https://github.com/hapijs/boom/issues/302) issue.
Given https://github.com/hapijs/hoek/pull/390 is now merged, can we get a new version of Boom published? Or is it awaiting [this](https://github.com/hapijs/boom/pull/304)? 

djs — 1:12 PM
Node was installing hoek@11.0.2 for boom. I fixed this by adding an explicit dependency to hoek@11.0.7 in my package.json.

### DIFF
--- a/designer/server/src/plugins/errorPage.ts
+++ b/designer/server/src/plugins/errorPage.ts
@@ -1,3 +1,4 @@
+import Boom from '@hapi/boom'
 import {
   type Request,
   type ResponseToolkit,
@@ -24,11 +25,10 @@ export default {
       server.ext('onPreResponse', (request: Request, h: ResponseToolkit) => {
         const response = request.response
 
-        if ('isBoom' in response && response.isBoom) {
+        if (Boom.isBoom(response)) {
           // An error was raised during
           // processing the request
           const statusCode = response.output.statusCode
-          const errorMessage = errorCodes.get(statusCode)
 
           if (statusCode === StatusCodes.NOT_FOUND.valueOf()) {
             request.logger.info(
@@ -40,14 +40,9 @@ export default {
               .code(statusCode)
           }
 
-          const logLevel =
-            statusCode === StatusCodes.NOT_FOUND.valueOf() ? 'info' : 'error'
-          request.logger[logLevel](
-            response,
-            statusCode === StatusCodes.NOT_FOUND.valueOf()
-              ? 'Resource not found'
-              : 'Unhandled error found'
-          )
+          request.logger.error(response, 'Unhandled error found')
+
+          const errorMessage = errorCodes.get(statusCode)
 
           if (errorMessage) {
             return h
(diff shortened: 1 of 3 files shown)


designer/server/src/plugins/errorPage.ts
```diff
@@ -40,21 +40,17 @@ export default {
...
           }
 
-          const logLevel =
-            statusCode === StatusCodes.NOT_FOUND.valueOf() ? 'info' : 'error'
-          request.logger[logLevel](
-            response,
-            statusCode === StatusCodes.NOT_FOUND.valueOf()
-              ? 'Resource not found'
-              : 'Unhandled error found'
-          )
```
alexluckett, line 50: good spot, wonder why we never spotted this unreachable code before!

alexluckett (approved):

PR #1375, opened 2026-04-09
Bump plugin version to v4.5.5

davidjamesstone:


alexluckett:
did you mean to open this PR? it's empty because of the revert.

davidjamesstone:
I did. It's a simple model bump but the first commit had build issues so I reverted. I'm just looking into why the bump caused the build failure - I think it's a package-lock.json problem.

davidjamesstone:
For reasons unknown, in bumping the engine version NPM changed the install location of `interactive-map` hence [this commit](https://github.com/DEFRA/forms-designer/pull/1375/changes/8b5c589db0b05bff8a77467f35759b500716f095). It should all build OK now.

PR #1376, opened 2026-04-10
Admin utilities to display form versions

alexluckett:
Adds a new Form Inspect developer tool to the admin area, gated behind a new FormsInspect scope (superadmin only). The tool lets administrators look up any form by ID or slug and inspect its state across several tabs.

# What's included

- Superadmin utility to view metadata, draft def, live def, versions and compare version history
- Added Scopes.FormsInspect = 'forms-inspect' to the model and routes
- Added jsondiffpatch as a dependency for structured JSON delta computation and server-side HTML rendering.

# Form Inspect routes (/admin/form-inspect)

```
Route	Description
GET /admin/form-inspect	Lookup form — radio + text input for ID or slug
GET /admin/form-inspect/{id}/metadata	View form metadata as formatted JSON
GET /admin/form-inspect/{id}/definition/live	View live form definition JSON
GET /admin/form-inspect/{id}/definition/draft	View draft form definition JSON
GET /admin/form-inspect/{id}/versions/{versionId}	View a specific version's definition JSON
GET /admin/form-inspect/{id}/versions/{from}..{to}	Render visual JSON diff between two versions
Version diffing with jsondiffpatch
```


designer/server/src/routes/admin/form-inspect.js
```diff
@@ -0,0 +1,527 @@
...
+   * @satisfies {ServerRoute<{ Payload: { type: string, id: string, slug: string } }>}
+   */
+  ({
+    method: 'POST',
+    path: ROUTE_FULL_PATH,
+    async handler(request, h) {
+      const { auth, payload } = request
+      const { token } = auth.credentials
+      const type = payload.type
+      const id = payload.id.trim()
```
jbarnsley10: Is it not better to do the `trim()` in the payload schema?

designer/server/src/routes/admin/form-inspect.js
```diff
@@ -0,0 +1,527 @@
...
+  ({
+    method: 'GET',
+    path: `${ROUTE_FULL_PATH}/{id}/definition/live`,
+    async handler(request, h) {
+      const { auth, params } = request
+      const { token } = auth.credentials
+      const { id } = params
+
+      let definition = null
+      try {
```
jbarnsley10, line 428: Why are you want to only throw for anything other than 404? Do you need the try/catch at all?

designer/server/src/routes/admin/form-inspect.js
```diff
@@ -0,0 +1,527 @@
...
+  ({
+    method: 'GET',
+    path: `${ROUTE_FULL_PATH}/{id}/definition/draft`,
+    async handler(request, h) {
+      const { auth, params } = request
+      const { token } = auth.credentials
+      const { id } = params
+
+      let definition = null
+      try {
```
jbarnsley10, line 477: Same as previous comment with try/catch

designer/server/src/routes/admin/form-inspect.js
```diff
@@ -0,0 +1,527 @@
...
+    },
+    options: {
+      auth: {
+        mode: 'required',
+        access: { entity: 'user', scope: [`+${Scopes.FormsInspect}`] }
+      }
+    }
+  })
+]
+
```
jbarnsley10, line 556: There are a few route methods with no param schema validation, but as you're checking in the code (or it's not critical), then that's fine

designer/server/src/views/admin/form-inspect-version-detail.njk
```diff
@@ -0,0 +1,32 @@
...
+      <h2 class="govuk-heading-m">Version {{ versionId }}</h2>
+      {{ appJsonDisplay(definition) }}
+    {% endcall %}
+
+  {% endcall %}
+{% endblock %}
+
+{% block mainEnd %}
+  {{ super() }}
+  <script type="module">
```
jbarnsley10, line 26: I like the fact you've re-used this :)

designer/server/src/views/admin/macros.njk
```diff
@@ -0,0 +1,24 @@
...
+      {% endfor %}
+    </ul>
+
+    <div class="govuk-tabs__panel">
+      {{ caller() }}
+    </div>
+  </div>
+{% endmacro %}
+
+{% macro appJsonDisplay(data) %}
```
jbarnsley10, line 17: Nice that this is now a macro. Is the DLQ stuff now using this macro or is that another PR?

jbarnsley10 (approved):
Excellent work. A couple of minor comments (and probs the 404 try/catches I now realise are because your page might display things like 'No draft versions for this form' etc), so probs ok.
Obvs forms-entitlement-api will need a model bump once this PR is mergd to pick up the new extra scope on superadmin.

designer/server/src/routes/admin/form-inspect.js
alexluckett: definitions becomes `FormDefinition | null`. If it's present, it becomes a def and we display it. If it's 404, it becomes null and we display a nice friendly message instead of letting it throw. If it's any other kind of error, we can't recover so we need to fail (throw).

designer/server/src/routes/admin/form-inspect.js
alexluckett: definitions becomes FormDefinition | null. If it's present, it becomes a def and we display it. If it's 404, it becomes null and we display a nice friendly message instead of letting it throw. If it's any other kind of error, we can't recover so we need to fail (throw).

designer/server/src/views/admin/macros.njk
alexluckett: Good point, I forgot to refactor that. I'll do that as a follow-up.

designer/server/src/routes/admin/form-inspect.js
alexluckett: good point

PR #1379, opened 2026-04-13
Add geospatial pattern link

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-856

- Add question details help text for Geospatial components
- Update placeholder image for Geospatial
- Bump engine to 4.5.6


jbarnsley10 (approved):
Approving anyway but just had a question if `scope: false` is correct for checking that a user is logged in but doesn't need any scopes

PR #1384, opened 2026-04-15
feat(DF-832): add conditional amounts and email prepopulation to PaymentField

mokhld:
- Add `conditionalAmounts` option to `PaymentFieldComponent` for condition-based payment amounts (first match wins, evaluated in order)
- Add `emailField` option to reference an `EmailAddressField` for GOV.UK Pay email prepopulation
- Update default amount minimum from £0.30 to £0 to support zero-amount bypass (no payment required when no conditions match)
- Conditional amounts enforce £0.30 minimum (GOV.UK Pay constraint)
- Update component-types fixture with empty `conditionalAmounts` default

### DIFF
--- a/model/src/form/form-definition/index.ts
+++ b/model/src/form/form-definition/index.ts
@@ -578,10 +578,10 @@ export const componentSchema = Joi.object<ComponentDef>()
       amount: Joi.when('type', {
         is: Joi.string().trim().valid(ComponentType.PaymentField).required(),
         then: Joi.number()
-          .min(0.3)
+          .min(0)
           .max(100000)
           .required()
-          .description('Payment amount in GBP (£0.30 - £100,000)')
+          .description('Payment amount in GBP (£0 - £100,000)')
       }).description('Payment amount - for PaymentField only'),
       description: Joi.when('type', {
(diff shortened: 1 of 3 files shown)


model/src/form/form-definition/index.ts
```diff
@@ -580,18 +578,46 @@ export const componentSchema = Joi.object<ComponentDef>()
...
+                .description('Amount in GBP when condition is true (min £0.30)')
+            })
+          )
+          .optional()
+          .description('Conditional payment amounts evaluated in order')
+      }).description('Conditional amounts - for PaymentField only'),
+      emailField: Joi.when('type', {
+        is: Joi.string().trim().valid(ComponentType.PaymentField).required(),
+        then: Joi.string()
+          .trim()
```
jbarnsley10, line 615: Does this need to be `email()` or are we happy that it will have been validated when the user entered it in the filed in the first place?

jbarnsley10 (approved):

model/src/form/form-definition/index.ts
mokhld: In the end, I decided not to do that because the email address field component already validates at input time, so by the time it reaches the form, it’s already valid.

PR #1390, opened 2026-04-16
Feature/df 827 min max checkboxes

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-827

### DIFF
--- a/designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
+++ b/designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
@@ -128,7 +128,11 @@ export function getAdditionalSchema(payload) {
     {
       key: 'min',
       sources: ['minLength', 'min', 'minFiles', 'minChecks'],
-      getValue: () => payload.minLength ?? payload.min ?? payload.minFiles ?? payload.minChecks,
+      getValue: () =>
+        payload.minLength ??
+        payload.min ??
+        payload.minFiles ??
+        payload.minChecks,
       shouldInclude: () =>
         payload.minLength ??
         isValueOrZero(payload.min) ??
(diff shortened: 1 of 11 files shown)


designer/server/src/models/forms/editor-v2/advanced-settings-schemas.js
```diff
@@ -11,6 +11,14 @@ const MIN_LENGTH_ERROR_MESSAGE =
   'Minimum length must be a positive whole number'
 const MAX_LENGTH_ERROR_MESSAGE =
   'Maximum length must be a positive whole number'
+const MIN_CHECKS_ERROR_MESSAGE =
+  'Minimum number of checkboxes must be a whole number greater than or equal to 2'
```
jbarnsley10, line 15: Are these error correct? Can you only set the values to 1 or 2 even if you have 10 checkboxes?

designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
```diff
@@ -127,24 +127,37 @@ export function getAdditionalSchema(payload) {
...
+        payload.min ??
+        payload.minFiles ??
+        payload.minChecks,
       shouldInclude: () =>
         payload.minLength ??
         isValueOrZero(payload.min) ??
-        isValueOrZero(payload.minFiles)
+        isValueOrZero(payload.minFiles) ??
+        isValueOrZero(payload.minChecks)
     },
```
alexluckett, line 141: This is probably due a refactor - bit janky how the code needs to "guess" where the config is coming from.

Would be much nicer to have something which is component-aware.

alexluckett (approved):

jbarnsley10 (approved):
Just one comment about error messages

designer/server/src/models/forms/editor-v2/advanced-settings-schemas.js
davidjamesstone: `greater than or equal to 2` so 2,3,4,5,6,7,8,9,10 would be valid

PR #1407, opened 2026-04-27
Feat/df 760 component usage 2

jbarnsley10:
Ticket [DF-736](https://eaflood.atlassian.net/browse/DF-736)

NOTE - entitlement-api needs a model bump from this PR to handle the new scope (for superadmin and admin users)

This UI handles many of the metrics covered in individual tickets, such as:

New forms created
Forms published
Form submissions
Table of metrics per form (including submissions)
Multiple tabs: 7 days, 30 days, all time
It does not yet handle the publishing metrics (so two tiles are missing from the page at the moment, and 'Days to publish' is left as '-').
image

[DF-736]: https://eaflood.atlassian.net/browse/DF-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


designer/server/src/models/admin/metrics-helper.js
```diff
@@ -0,0 +1,358 @@
...
+      ([_aKey, aVal], [_bKey, bVal]) => bVal - aVal
+    )
+  )
+
+  return Array.from(featuresSorted).map((f) => ({
+    featureName: f[0],
+    formsUsing: f[1],
+    percentage: `${((f[1] / totalForms) * 100).toFixed(1)}%`
+  }))
+}
```
alexluckett: since you don't appear to need the outer map, this could be simplified

`[...formsUsing.entries(...)].sort(...).map(...)`?

designer/server/src/models/admin/metrics-helper.js
```diff
@@ -0,0 +1,358 @@
...
+
+/**
+ * @param {string} metricName
+ */
+function mapFormStructureName(metricName) {
+  if (metricName === 'questionTypes') {
+    return 'Question types per form'
+  }
+  return `${metricName.substring(0, 1).toUpperCase()}${metricName.substring(1)} per form`
```
alexluckett: blows my mind that JS doesn't have a capitalise function built-in. Another Python win! 😜

alexluckett: bit of an odd convention that we're passing in some kind of ID with `questionTypes` but other times it's a more human readable metric.

I'd be consistent and keep either an ID or a human-readable metric name as the input, rather than mixing concerns.

just checking and you're actually already doing this for `tilePeriodNames`

or just ditch the mapper altogether and have a static config, e.g.

somewhere else:

```
const metricNames = {
    "questionTypes": "question types",
    "numberOfLiveForms": "number of live forms"
}
```

```
function capitalise(str) {
    return ${str.substring(0, 1).toUpperCase()}${str.substring(1)}
}

edit: capitalise isn't neccessary if we're doing hardcoded strings. I overcomplicated it. My point was to show that it's a generic utility that can be pulled out.

function mapFormStructureName(metricId) {
    const metricTitle = capitalise(metricNames.get(metricId))

    return metricTitle
}
```

jbarnsley10: Used a record lookup to be consistent

alexluckett (approved):

PR #1419, opened 2026-05-05
feat/df-1006: Resubmit for all DLQs

jbarnsley10:
- Resubmit for all DLQs
- Tuned listing of DLQs

### DIFF
--- a/designer/server/src/lib/dead-letter-queue.js
+++ b/designer/server/src/lib/dead-letter-queue.js
@@ -56,6 +56,7 @@ export async function getDeadLetterQueueMessages(dlq, token, options) {
 
   const requestUrl = new URL(`./admin/deadletter${qualifier}/view`, endpoint)
 
+  // Allow zero as valid parameter value
   if (options?.visibilityTimeout !== undefined) {
     requestUrl.searchParams.set(
       'visibilityTimeout',
@@ -63,6 +64,7 @@ export async function getDeadLetterQueueMessages(dlq, token, options) {
     )
   }
 
+  // Allow zero as valid parameter value
   if (options?.waitTimeSeconds !== undefined) {
     requestUrl.searchParams.set(
       'waitTimeSeconds',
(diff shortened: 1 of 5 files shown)


designer/server/src/lib/dead-letter-queue.js
```diff
@@ -56,14 +56,14 @@ export async function getDeadLetterQueueMessages(dlq, token, options) {
 
   const requestUrl = new URL(`./admin/deadletter${qualifier}/view`, endpoint)
 
-  if (options?.visibilityTimeout) {
```
alexluckett, line 59: the original apporoach seemed fine?

designer/server/src/routes/admin/dead-letter-queue-helper.js
```diff
@@ -1,11 +1,34 @@
...
+    return messageSchema
+  }
+  if (
+    dlq === DeadLetterQueues.SubmissionsApiSaveAndExit.toString() ||
+    dlq === DeadLetterQueues.SubmissionsApiFormSubmissions.toString()
+  ) {
+    return submissionMessageSchema
+  }
+  return formAdapterSubmissionMessagePayloadSchema
+}
```
alexluckett, line 32: An object/map is much cleaner for mapping - no logic statements neccessary. 

```
const queueToSchema = {
  [DeadLetterQueues.SubmissionsApiSaveAndExit]: submissionMessageSchema,
  [DeadLetterQueues.SubmissionsApiFormSubmissions]: submissionMessageSchema,
  [DeadLetterQueues.AuditApi]: messageSchema,
  [DeadLetterQueues.TheNameOftheOtherQueue]: formAdapterSubmissionMessagePayloadSchema,
}

export function getCorrectMessageSchema(dlq) {
  const schema = queueToSchema[dlq]

  if (!schema) {
    throw new Error(`Unknown queue: "${dlq}"`) // always fail safe, we don't want to silently fall back on the wrong schema
  }

  return schema
}
```

designer/server/src/lib/dead-letter-queue.js
jbarnsley10: The original didn't handle a timeout value of zero. I want to be able to override with a value of zero for the count listings of the queues.

alexluckett: gotcha

alexluckett (approved):

PR #1420, opened 2026-05-05
feat/df-922: Metrics download

jbarnsley10:
Allow download of metrics data
According to ticket [DF-992](https://eaflood.atlassian.net/browse/DF-992), it should only download submissions counts for LIVE (non-preview) submissions

[DF-992]: https://eaflood.atlassian.net/browse/DF-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


designer/server/src/messaging/mappers/events.js
```diff
@@ -167,6 +167,30 @@ function excelGenerationBaseMapper(data, user, type) {
   }
 }
 
+/**
+ * Base mapper for direct downloads
+ * @param {AuditUser} user
+ * @param {AuditEventMessageType.PLATFORM_METRICS_DOWNLOAD_REQUESTED} type
+ * @returns {PlatformMetricsDownloadRequestedMessage}
+ */
+function directDownloadBaseMapper(user, type) {
```
alexluckett, line 176: Bit late in the game for me to be saying this, but auditing metrics downloads seems like overkill to me. Happy to leave it, just an observation really.

designer/server/src/models/admin/metrics.js
```diff
@@ -156,6 +157,50 @@ export function combineModel(combinedElement, liveElement, typeKeyName) {
...
+
+  const contentOutput = /** @type {string[]} */ ([
+    '"Form name","Form URL","Live submissions"'
+  ])
+  formsSorted.forEach((ov) => {
+    const summaryMetrics = /** @type {{ name: string, slug: string }} */ (
+      ov.summaryMetrics
+    )
+    const escapedFormName = summaryMetrics.name
+      .replaceAll('"', '')
```
alexluckett: Is this actually possible?

alexluckett: I also find that using CSV libraries are better than doing this manually, as they deal with escaping directly rather than us needing to deal with it directly.

We use csv-stringify elsewhere: https://github.com/DEFRA/forms-submission-api/blob/636233f65b2e6d3b9658d28e28ffbd7a48c84d5d/src/services/utils.js#L13

designer/server/src/models/admin/metrics.js
```diff
@@ -156,6 +157,50 @@ export function combineModel(combinedElement, liveElement, typeKeyName) {
...
+    const count = metrics.totals.liveSubmissions
+      ? /** @type {number | undefined} */ (
+          metrics.totals.liveSubmissions[form.formId]
+        )
+      : 0
+    form.submissionsCount = count ?? 0
+  })
+
+  // Sort forms by name then status
+  const formsSorted = liveOnly.toSorted((a, b) => {
```
Scullyon, line 205: Status is now always live, so do you need to include status in the comparison anymore?

designer/server/src/models/admin/metrics.js
jbarnsley10: Now uses csv-stringify

jbarnsley10: Uses csv-stringify now and that handles the escaping (if necessary)

alexluckett (approved):

PR #1422, opened 2026-05-06
feat(DF-830): conditional payment amounts editor

mokhld:
Adds a conditional payment amounts editor to PaymentField so editors can configure different charges per matched condition. All ticket flows verified.

### DIFF
--- a/designer/server/src/lib/editor.js
+++ b/designer/server/src/lib/editor.js
@@ -712,6 +712,52 @@ export async function deleteCondition(formId, token, conditionId) {
 }
 
 /**
+ * @param {string} formId
+ * @param {string} token
+ * @param {FormDefinition} definition
+ * @param {string} pageId
+ * @param {string} componentId
+ * @param {string} conditionId
+ */
+export async function removePaymentConditionalAmountReference(
+  formId,
+  token,
+  definition,
+  pageId,
+  componentId,
+  conditionId
+) {
+  const page = getPageFromDefinition(definition, pageId)
+  const components = hasComponents(page) ? page.components : []
+  const component =
+    /** @type {Partial<ComponentDef> & { options?: { conditionalAmounts?: Array<{ amount: number, condition: string }> } } | undefined} */ (
+      components.find((c) => c.id === componentId)
+    )
+  if (!component) {
+    return
+  }
+  const existing = component.options?.conditionalAmounts ?? []
+  const filtered = existing.filter((entry) => entry.condition !== conditionId)
+  if (filtered.length === existing.length) {
+    return
+  }
+  const updated = {
+    ...component,
+    options: {
+      ...(component.options ?? {}),
+      conditionalAmounts: filtered
+    }
+  }
+  await putJson(
+    buildRequestUrl(formId, `pages/${pageId}/components/${componentId}`),
+    {
+      payload: updated,
+      ...getHeaders(token)
+    }
+  )
+}
+
+/**
  * @typedef {object} UpdateSectionsResponse
  * @property {string} id - Form ID
  * @property {Section[]} sections - Updated sections with IDs
--- a/designer/server/src/lib/editor.js
+++ b/designer/server/src/lib/editor.js
@@ -712,52 +712,6 @@ export async function deleteCondition(formId, token, conditionId) {
 }
 
 /**
- * @param {string} formId
- * @param {string} token
- * @param {FormDefinition} definition
- * @param {string} pageId
- * @param {string} componentId
- * @param {string} conditionId
- */
-export async function removePaymentConditionalAmountReference(
-  formId,
-  token,
-  definition,
-  pageId,
-  componentId,
-  conditionId
-) {
-  const page = getPageFromDefinition(definition, pageId)
-  const components = hasComponents(page) ? page.components : []
-  const component =
-    /** @type {Partial<ComponentDef> & { options?: { conditionalAmounts?: Array<{ amount: number, condition: string }> } } | undefined} */ (
-      components.find((c) => c.id === componentId)
-    )
-  if (!component) {
-    return
-  }
-  const existing = component.options?.conditionalAmounts ?? []
-  const filtered = existing.filter((entry) => entry.condition !== conditionId)
-  if (filtered.length === existing.length) {
-    return
-  }
-  const updated = {
-    ...component,
-    options: {
-      ...(component.options ?? {}),
-      conditionalAmounts: filtered
-    }
-  }
-  await putJson(
-    buildRequestUrl(formId, `pages/${pageId}/components/${componentId}`),
-    {
-      payload: updated,
-      ...getHeaders(token)
-    }
-  )
-}
-
-/**
  * @typedef {object} UpdateSectionsResponse
  * @property {string} id - Form ID
  * @property {Section[]} sections - Updated sections with IDs
--- a/designer/server/src/lib/payment-conditional-amount-helpers.js
+++ b/designer/server/src/lib/payment-conditional-amount-helpers.js
@@ -1,5 +1,7 @@
 import { randomUUID } from 'node:crypto'
 
+import { ComponentType } from '@defra/forms-model'
+
 /**
  * @param {ConditionalAmountState[] | undefined} items
  * @param {string} id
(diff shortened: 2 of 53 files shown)


designer/server/src/routes/forms/editor-v2/question-details.js
```diff
@@ -189,22 +228,64 @@ export function overrideStateIfJsEnabled(request) {
...
+  if (questionDetails.type === ComponentType.PaymentField) {
+    const hydrated = hydrateConditionalAmountsFromComponent(
+      getComponentFromDefinition(definition, pageId, questionId),
+      state
+    )
+    if (hydrated !== state) {
+      setQuestionSessionState(yar, stateId, hydrated)
+    }
+  }
+  handleSaveConditionalAmount(request, stateId)
```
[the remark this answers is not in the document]
mokhld: This will change the intended behaviour.

designer/server/src/routes/forms/editor-v2/condition-delete.js
```diff
@@ -63,6 +67,23 @@ export default [
...
+        for (const ref of paymentFields) {
+          await removePaymentConditionalAmountReference(
+            formId,
+            token,
+            definition,
+            ref.pageId,
+            ref.componentId,
+            conditionId
+          )
+        }
```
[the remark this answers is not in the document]
mokhld: I think this is a premature optimisation.

designer/server/src/routes/forms/editor-v2/question-details.js
```diff
@@ -170,6 +182,33 @@ export function missingPageTitleForMultipleQuestions(questionId, page) {
   return questionId === 'new' && requiresPageTitle(page)
 }
 
+/**
+ * @param {QuestionSessionState} existing
+ * @param {string | undefined} questionType
+ * @param {string | undefined} listItemsData
+ * @returns {QuestionSessionState}
+ */
+function buildOverridenState(existing, questionType, listItemsData) {
```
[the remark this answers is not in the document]
mokhld: Good catch.

designer/server/src/models/forms/editor-v2/pages.js
```diff
@@ -115,6 +115,22 @@ export function mapMarkdown(component, isSummary) {
...
+  const amount = component.options.amount
+  if (amount > 0) {
+    return formatCurrency(amount)
+  }
+  const conditional = component.options.conditionalAmounts ?? []
+  if (conditional.length > 0) {
+    return formatCurrency(conditional[0].amount)
+  }
+  return formatCurrency(0)
+}
```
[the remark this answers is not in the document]
mokhld: Minor but valid. Will fix.

designer/server/src/lib/editor.js
```diff
@@ -711,6 +711,52 @@ export async function deleteCondition(formId, token, conditionId) {
...
 
+/**
+ * @param {string} formId
+ * @param {string} token
+ * @param {FormDefinition} definition
+ * @param {string} pageId
+ * @param {string} componentId
+ * @param {string} conditionId
+ */
+export async function removePaymentConditionalAmountReference(
```
alexluckett: usually lib is a lightweight wrapper around the API, this feels like more of a service layer operation?

designer/server/src/lib/payment-conditional-amount-helpers.js
```diff
@@ -126,6 +133,19 @@ export function hydrateConditionalAmountsFromComponent(component, state) {
 }
 
 /**
- * @import { ComponentDef } from '@defra/forms-model'
+ * @param {PaymentFieldComponent} component
+ * @returns {number}
+ */
+export function getPaymentDisplayAmount(component) {
```
alexluckett: ditto RE service point

designer/server/src/lib/payment-conditional-amount-helpers.js
```diff
@@ -126,6 +133,19 @@ export function hydrateConditionalAmountsFromComponent(component, state) {
...
+  const amount = component.options.amount
+  if (amount > 0) {
+    return amount
+  }
+  const conditional = component.options.conditionalAmounts ?? []
+  return conditional.length > 0 ? conditional[0].amount : 0
```
alexluckett: Why would we always take [0] and display that?

designer/server/src/models/forms/editor-v2/base-settings-fields.js
```diff
@@ -138,13 +138,9 @@ export const baseSchema = Joi.object().keys({
   enhancedAction: questionDetailsFullSchema.enhancedActionSchema,
   radioId: questionDetailsFullSchema.radioIdSchema,
   radioText: questionDetailsFullSchema.radioTextSchema.when('enhancedAction', {
-    is: Joi.exist(),
-    then: Joi.string().when('enhancedAction', {
-      is: Joi.string().valid('add-item', 're-order'),
-      then: Joi.string().optional().allow(''),
-      otherwise: Joi.string().trim().required().messages({
```
alexluckett, line 145: was this not needed anymore?

designer/server/src/models/forms/editor-v2/base-settings-fields.js
```diff
@@ -346,9 +342,11 @@ function getPaymentAmount(questionFields) {
...
     questionFields
   )
-  return paymentField?.options.amount
-    ? paymentField.options.amount.toFixed(2)
-    : undefined
+  const amount = paymentField?.options.amount
+  if (typeof amount !== 'number') {
+    return '0'
+  }
+  return amount === 0 ? '0' : amount.toFixed(2)
```
alexluckett, line 353: differing logic from `getPaymentDisplayAmount`? intentional?

designer/server/src/routes/forms/editor-v2/payment-conditional-amount-actions.js
```diff
@@ -95,6 +111,56 @@ export function handleSaveConditionalAmount(request, stateId) {
...
+  const duplicate = items.some(
+    (item) => item.id !== editingId && item.condition === value.condition
+  )
+  if (duplicate) {
+    return buildConditionError(
+      'You already have a payment amount for this condition'
+    )
+  }
+  return null
+}
```
alexluckett, line 150: could you do a demo of this? struggling to piece together how it works from the code

designer/server/src/routes/forms/editor-v2/condition-delete.js
```diff
@@ -63,6 +67,23 @@ export default [
       const formId = metadata.id
 
       try {
+        const definition = await forms.getDraftFormDefinition(formId, token)
+        const { paymentFields } = findConditionReferences(
+          definition,
+          conditionId
+        )
```
alexluckett, line 71: good catch

designer/server/src/lib/payment-conditional-amount-helpers.js
mokhld: I see what you mean. Although this is a pure function, I'm not really sure if it can be considered a service operation, but I'll move it anyway, as it's pretty simple.

designer/server/src/lib/payment-conditional-amount-helpers.js
mokhld: So this matches [the runtime preview model's behaviour,](https://github.com/DEFRA/forms-designer/blob/28a6261135e9f666afaaa32979aee0f82fe6ac4a/model/src/form/form-editor/preview/payment.js#L153) which also takes `[0]`. My logic is that at design time, we don't know which condition will match, so we show a representative figure rather than showing £0. We could show a range like from £X or up to £Y, but I think that's more for product call?

designer/server/src/models/forms/editor-v2/base-settings-fields.js
mokhld: Still needed, but narrower scope. The old gate required `radioText` for any `enhancedAction` NOT in `[add-item, re-order]` Which is also now capturing the new conditional amount actions and would otherwise block them with the misleading "enter item text" error. So now my logic is to require `radioText` only when `enhancedAction === 'save-item'` which is  the only action that actually saves radios list item text

designer/server/src/models/forms/editor-v2/base-settings-fields.js
mokhld: Yep, this is intentional as there are different domains:
`getPaymentAmount` - Prefills the editors payment amount base input, and it will default to 0 for empty. There's no fallback to the conditional, as falling back would be wrong because the user is editing the base value. 
`getPaymentDisplayAmount` This is for the end user display, and this will fall back to the first conditional when the base is 0, so the end users don't see "£0" on a payment required form.  I added a JSDoc comment to clarify as well.

alexluckett (approved):

PR #1431, opened 2026-05-11
feat(DF-952): align designer with new map guidance

mokhld:
Bumps engine to ^4.11.3, switches the four location-field default instructions to the shared 4-step map pattern, and refreshes the location preview image.


.gstack/browse-startup-error.log
```diff
@@ -0,0 +1,21 @@
+2026-05-11T10:39:39.551Z launch: Executable doesn't exist at /Users/abusafiyyah/Library/Caches/ms-playwright/chromium_headless_shell-1208/chrome-headless-shell-mac-arm64/chrome-headless-shell
```
Scullyon: This supposed to be here?

mokhld: Ah, nope. Removing.

Scullyon (approved):

davidjamesstone (commented):
I think that placeholder JPEG might need trimming/clipping to the grey border.

mokhld:
> I think that placeholder JPEG might need trimming/clipping to the grey border.

I initially did that. The background is white, so it doesn't necessarily look out of place, but I'll redo that.

mokhld:
> > I think that placeholder JPEG might need trimming/clipping to the grey border.
> 
> I initially did that. The background is white, so it doesn't necessarily look out of place, but I'll redo that.

That's been updated @davidjamesstone

davidjamesstone:
> > I think that placeholder JPEG might need trimming/clipping to the grey border.
> 
> I initially did that. The background is white, so it doesn't necessarily look out of place, but I'll redo that.

Thanks Mo. I remembered I had to clip it as it looked a little bit out of alignment :+1:

PR #1443, opened 2026-05-14
feat(DF-830): add paymentConditionalAmounts field to editor and tests

mokhld:
- Introduced paymentConditionalAmounts to the paymentFields array in base-settings-fields.js.
- Added tests to ensure paymentConditionalAmounts is correctly positioned and utilizes the custom template in base-settings-fields.test.js.


nayij (approved):
LGTM

PR #1449, opened 2026-05-18
404 on Responses Page

Scullyon:
Fixing issue with 'Page not found' on responses page when draft does not exist. This issue was caused by an attempt to build navigation from the draft version of the form which may not exist.

### DIFF
--- a/designer/server/src/lib/forms.js
+++ b/designer/server/src/lib/forms.js
@@ -125,6 +125,18 @@ export async function getLiveFormDefinition(id, token) {
 }
 
 /**
+ * Get form definition, preferring draft over live
+ * @param {FormMetadata} metadata
+ * @param {string} token
+ */
+export function getFormDefinition(metadata, token) {
+  if (metadata.draft) {
+    return getDraftFormDefinition(metadata.id, token)
+  }
+  return getLiveFormDefinition(metadata.id, token)
+}
+
+/**
  * Retrieves a form definition version from the form manager for a given id
  * @param {string} id - the id of the form
  * @param {number} versionNumber - the version of the form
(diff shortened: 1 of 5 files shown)


designer/server/src/lib/forms.js
```diff
@@ -124,6 +124,18 @@ export async function getLiveFormDefinition(id, token) {
...
+ * Get form definition, preferring draft over live
+ * @param {FormMetadata} metadata
+ * @param {string} token
+ */
+export function getFormDefinition(metadata, token) {
+  if (metadata.draft) {
+    return getDraftFormDefinition(metadata.id, token)
+  }
+  return getLiveFormDefinition(metadata.id, token)
+}
```
alexluckett: A bit dangerous, from the method signature alone it's unclear what will be returned. Might be a situation where someone accidentally uses a draft because the form happens to have one. 

a nicer approach is what we do in [forms-notify-listener](https://github.com/DEFRA/forms-notify-listener/blob/fff9594139ab48a9ca94e8152c13c121ac8804c8/src/lib/manager.js#L14) where we take `formStatus` in as a param which becomes: `getFormDefinition(metadata, token, FormStatus.Draft)` or `getFormDefinition(metadata, token, FormStatus.Live)`

Scullyon: We should have a library (not models) for this shared functionality stuff. As an aside, arguably the function you've linked to should ensure that `formStatus` is supplied via validation as it's not marked as an optional parameter in doc-comment.

Will update the method name for now as we already have two existing methods that get live or draft.

Scullyon: Have updated this.

davidjamesstone (approved):

PR #1455, opened 2026-05-19
fix/df-1104: Hides metric filter by default + handles seach text with spaces

jbarnsley10:
- Fix to handle metrics search text that may contain spaces
- Hides filter by default, unless it contains filter values

### DIFF
--- a/designer/server/src/routes/admin/form-metrics.js
+++ b/designer/server/src/routes/admin/form-metrics.js
@@ -8,7 +8,11 @@ import Joi from 'joi'
 import { mapUserForAudit } from '~/src/common/helpers/auth/user-helper.js'
 import { logger } from '~/src/common/helpers/logging/logger.js'
 import { buildAdminNavigation } from '~/src/common/nunjucks/context/build-navigation.js'
-import { getMetrics, regenerateMetrics } from '~/src/lib/metrics.js'
+import {
+  MetricsFilterFields,
+  getMetrics,
+  regenerateMetrics
+} from '~/src/lib/metrics.js'
 import { publishPlatformMetricsDownloadRequestedEvent } from '~/src/messaging/publish.js'
 import {
   getLiveMetricsAsCsv,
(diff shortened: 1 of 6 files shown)


designer/server/src/routes/admin/form-metrics.js
```diff
@@ -47,21 +47,24 @@ export function buildQueryFromPayload(payload) {
   }
 
   const params = new URLSearchParams()
-  if (payload.showFilter === 'N') {
-    params.set('showFilter', 'N')
-  }
   if (payload.searchText) {
-    params.set('searchText', encodeURI(payload.searchText))
+    params.set('searchText', encodeURI(payload.searchText.trim()))
+    params.set('showFilter', 'Y')
```
Scullyon: Minor point: there are a lot of magic strings all over the place. Would it be better to factor them out into a central location?

Scullyon (dismissed):
.

designer/server/src/routes/admin/form-metrics.js
```diff
@@ -47,21 +47,24 @@ export function buildQueryFromPayload(payload) {
   }
 
   const params = new URLSearchParams()
-  if (payload.showFilter === 'N') {
-    params.set('showFilter', 'N')
-  }
   if (payload.searchText) {
-    params.set('searchText', encodeURI(payload.searchText))
+    params.set('searchText', encodeURI(payload.searchText.trim()))
```
Scullyon: The `set` method already performs encoding. So this is double encoding params?

jbarnsley10: I had missed the fact I was double-encoding. Fixed now

designer/server/src/routes/admin/form-metrics.js
jbarnsley10: Changed

Scullyon (approved):

PR #1460, opened 2026-05-21
DF-925: Download all submission files

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-925

image

image

image

### DIFF
--- a/designer/client/src/javascripts/download.js
+++ b/designer/client/src/javascripts/download.js
@@ -105,10 +105,12 @@ function onDownloadFinished(fileId) {
  * @param {string} email - the email address of the user
  */
 async function downloadAll(email) {
-  const links = document.querySelectorAll('.govuk-summary-list__actions a')
+  const links = /** @type {NodeListOf<HTMLAnchorElement>} */ (
+    document.querySelectorAll('.govuk-summary-list__actions a')
+  )
   const files = Array.from(links).map((link) => ({
     href: /** @type {string} */ (link.getAttribute('href')),
-    fileId: /** @type {string} */ (link.getAttribute('data-fileid'))
+    fileId: /** @type {string} */ (link.dataset['data-fileid'])
   }))
 
   await downloadAllFiles(files, email)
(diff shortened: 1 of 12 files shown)


designer/client/src/javascripts/download.js
```diff
@@ -0,0 +1,219 @@
...
+  if (response.ok) {
+    const { url, fileName } = await response.json()
+    await download(url, fileName)
+  } else {
+    const forbiddenCode = 403
+    const next = `/file-download/${fileId}`
+
+    // Redirect to email entry page
+    // If the response was forbibben then include the reference
+    // number to they come back to the download all page afterward
```
jbarnsley10: typo 'to' -> 'so' ?

davidjamesstone: Resolved with https://github.com/DEFRA/forms-designer/pull/1460/commits/9d02bd4b02482c2e13af3dacad209a609e31ac31

designer/server/src/models/file/file.js
```diff
@@ -30,5 +33,85 @@ export function fileViewModel(email, validation) {
...
+}
+
+/**
+ * Find component by name
+ * @param {FormDefinition} definition
+ * @param {string} componentName
+ * @returns {ComponentDef}
+ * @throws {Error} If component is not found
+ */
+export function getComponent(definition, componentName) {
```
jbarnsley10, line 98: We might want this method in the model common helpers, but because it seems quite specific to the download functionality, happy for it to remain here.

davidjamesstone: I was surprised we didn't have it already. We have things for getComponentFromPage and getPageFromDef.

I could stick it in `designer/server/src/lib/utils.js` or maybe `model/src/components/helpers.ts` if you think it's useful beyond download?

jbarnsley10: On second thoughts, I'd leave it where it is, as can't think of any other areas that would use it. Up to you though

jbarnsley10 (approved):
Great work.
Just the issue of localhost vs host.docker.internal when running locally but that can be perhaps a spearate PR

davidjamesstone:
> Great work. Just the issue of localhost vs host.docker.internal when running locally but that can be perhaps a spearate PR

Fixed in https://github.com/DEFRA/forms-designer/pull/1460/commits/1d52704c60480c0752f6a233d3dbdbab1f2cd036

PR #1464, opened 2026-05-26
feat/df-892: Deletes payment key as appropriate

jbarnsley10:
- Deletes TEST payment key if deleting a draft page/question (and was a payment question/page)
- Deletes LIVE payment key if going live and the new live doesn't contain a payment question any more

Ticket [DF-892](https://eaflood.atlassian.net/browse/DF-892)

[DF-892]: https://eaflood.atlassian.net/browse/DF-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/designer/server/src/lib/secrets.js
+++ b/designer/server/src/lib/secrets.js
@@ -191,19 +191,16 @@ export async function savePaymentSecrets(
 
 /**
  * @param {string} formId
- * @param {boolean} isLive
+ * @param {string} secretName
  * @param {string} token
  */
-export async function deletePaymentSecret(formId, isLive, token) {
-  const key = isLive ? PAYMENT_LIVE_API_KEY_PENDING : PAYMENT_TEST_API_KEY
+export async function deletePaymentSecret(formId, secretName, token) {
   const { response } = await delJson(
-    buildRequestUrl(formId, key),
+    buildRequestUrl(formId, secretName),
     getHeaders(token)
   )
   if (response.statusCode !== StatusCodes.OK) {
-    throw new Error(
-      `Failed to delete ${isLive ? 'LIVE' : 'TEST'} Payment API key`
-    )
+    throw new Error(`Failed to delete ${secretName} Payment API key`)
   }
 }
 
--- a/designer/server/src/lib/secrets.js
+++ b/designer/server/src/lib/secrets.js
@@ -3,7 +3,7 @@ import Boom from '@hapi/boom'
 import { StatusCodes } from 'http-status-codes'
 
 import config from '~/src/config.js'
-import { getJson, postJson } from '~/src/lib/fetch.js'
+import { delJson, getJson, postJson } from '~/src/lib/fetch.js'
 import { getLiveFormDefinition } from '~/src/lib/forms.js'
 import { getHeaders } from '~/src/lib/utils.js'
 
@@ -187,6 +187,24 @@ export async function savePaymentSecrets(
     }
   }
   return ''
+}
+
+/**
+ * @param {string} formId
+ * @param {boolean} isLive
+ * @param {string} token
+ */
+export async function deletePaymentSecret(formId, isLive, token) {
+  const key = isLive ? PAYMENT_LIVE_API_KEY_PENDING : PAYMENT_TEST_API_KEY
+  const { response } = await delJson(
+    buildRequestUrl(formId, key),
+    getHeaders(token)
+  )
+  if (response.statusCode !== StatusCodes.OK) {
+    throw new Error(
+      `Failed to delete ${isLive ? 'LIVE' : 'TEST'} Payment API key`
+    )
+  }
 }
 
 /**
--- a/designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
+++ b/designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
@@ -13,6 +13,7 @@ import { deletePage, deleteQuestion } from '~/src/lib/editor.js'
 import { isInvalidFormErrorType } from '~/src/lib/error-boom-helper.js'
 import * as forms from '~/src/lib/forms.js'
 import {
+  PAYMENT_LIVE_API_KEY_PENDING,
   PAYMENT_TEST_API_KEY,
   deletePaymentSecret,
   existsSecret
@@ -106,11 +107,8 @@ export default [
         }
 
         if (isPayment) {
-          // Delete TEST payment API key (if present)
-          const exists = await existsSecret(formId, PAYMENT_TEST_API_KEY, token)
-          if (exists.exists) {
-            await deletePaymentSecret(formId, false, token)
-          }
+          // Handle deletion of payment API keys
+          await cleanupPaymentKeys(formId, metadata, token)
         }
 
         // Redirect POST to GET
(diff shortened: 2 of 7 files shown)


designer/server/src/lib/secrets.js
```diff
@@ -189,6 +189,21 @@ export async function savePaymentSecrets(
   return ''
 }
 
+/**
+ * @param {string} formId
+ * @param {string} secretName
+ * @param {string} token
+ */
+export async function deletePaymentSecret(formId, secretName, token) {
+  const { response } = await delJson(
```
Scullyon, line 198: Appreciate this is more of a comment aimed at `fetch.request` function that this actually uses, but is pertinent here: shouldn't we have some kind of exponential back-off retry logic in here to guard against transient errors?

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
```diff
@@ -84,13 +96,21 @@ export default [
...
+        const isPayment = isPaymentPage(page)
+
         // If only one (non-guidance question) on the page, 'deleting the question' becomes 'deleting the page'
         if (questionId && shouldDeleteQuestionOnly(pageId, definition)) {
           await deleteQuestion(formId, token, pageId, questionId, definition)
         } else {
           await deletePage(formId, token, pageId, definition)
         }
 
+        if (isPayment) {
```
Scullyon, line 109: It's true that we can only have one payment page at the moment. Will this always be true? Is there any need to check for other payment questions? Or is that something to tackle as part of multiple payment questions being allowed (if that ever made sense?

designer/server/src/routes/forms/form-lifecycle.js
```diff
@@ -63,6 +72,15 @@ export default [
       try {
         await forms.makeDraftFormLive(form.id, token)
 
+        // Delete payment secret if no longer in live form
+        const exists = await existsSecret(form.id, PAYMENT_LIVE_API_KEY, token)
```
Scullyon: liveApiPaymentKey?

designer/server/src/routes/forms/form-lifecycle.js
```diff
@@ -63,6 +72,15 @@ export default [
...
+        if (exists.exists) {
+          const definition = await forms.getLiveFormDefinition(form.id, token)
```
Scullyon, line 78: Is there a check elsewhere that does the opposite of this (ie checks that if there's a payment question, that there's a live payment API key)?

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
```diff
@@ -84,13 +96,21 @@ export default [
       const definition = await forms.getDraftFormDefinition(formId, token)
 
       try {
+        const page = getPageFromDefinition(definition, pageId)
+        const isPayment = isPaymentPage(page)
```
Scullyon, line 100: Technically, this may not currently be an issue, but if you could have more than one question on a page with a payment question, then someone may be deleting the non-payment question. So it's more than just whether this page is a payment page if the whole page is not being deleted? It's also whether the question being deleted is the payment question?

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
```diff
@@ -84,13 +96,21 @@ export default [
...
         // If only one (non-guidance question) on the page, 'deleting the question' becomes 'deleting the page'
         if (questionId && shouldDeleteQuestionOnly(pageId, definition)) {
           await deleteQuestion(formId, token, pageId, questionId, definition)
         } else {
           await deletePage(formId, token, pageId, definition)
         }
 
+        if (isPayment) {
+          // Handle deletion of payment API keys
+          await cleanupPaymentKeys(formId, token)
```
Scullyon, line 111: What happens if this bit fails, but the question and/or page has successfully been deleted?

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
```diff
@@ -133,6 +153,28 @@ export default [
   })
 ]
 
+/**
+ * @param {string} formId
+ * @param {string} token
+ */
+async function cleanupPaymentKeys(formId, token) {
+  // Delete TEST payment API key (if present)
+  const exists = await existsSecret(formId, PAYMENT_TEST_API_KEY, token)
```
Scullyon: You could run these two in parallel to speed things up (the TEST and LIVE_PENDING operations).

designer/server/src/lib/secrets.js
jbarnsley10: I think we've got major problems if a simple fetch fails to communicate with another container. I wouldn't imagine implementing a retry would add much value TBH. We just throw and let the user retry if they want.

jbarnsley10: Maybe one to suggest to Alex

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
jbarnsley10: At the moment, we can only have one payment question per page, and only ever one payment question in the entire form. If that changes, we will need to rework some of this, but I think it's unlikely we'd ever have more than one payment question in a form due to the implementation of conditional payments.

designer/server/src/routes/forms/form-lifecycle.js
jbarnsley10: There are 3 constants which all start with 'PAYMENT_' for the API payment keys. They define the names of the appropriate secrets.
Are you suggesting they should be called 'LIVE_API_PAYMENT_KEY' and 'TEST_API_PAYMENT_KEY' etc?

designer/server/src/routes/forms/form-lifecycle.js
jbarnsley10: Yes, the go-live process enforces live API keys if you have a payment question.

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
jbarnsley10: As above really. When deleting a payment question, it always (currently) ends up deleting the page as only ever one question allowed on that page.

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
jbarnsley10: It throws. Not too much we can do, I don't think, since the update of page definition is a separate API call to the deleting of secrets.

designer/server/src/routes/forms/editor-v2/question-or-page-delete.js
jbarnsley10: This is true. I'll make a change to do that.

designer/server/src/lib/secrets.js
Scullyon: Yeah, I think so. It should be the default for all comms because transient issues do arise that fix themselves in less than a second. I'll add it to my list to raise as it should probably be everywhere.

designer/server/src/routes/forms/form-lifecycle.js
Scullyon: Sorry, the selected variable didn't highlight as part of the comment. 

I mean the `const exists` should perhaps be `const liveApiPaymentKey`, so that later on there isn't `exists.exists`.

Scullyon (approved):

designer/server/src/routes/forms/form-lifecycle.js
jbarnsley10: Sure. I've changed the var name and pushed

PR #1470, opened 2026-05-28
Email responses to auth user email not the shared inbox

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-1030


jbarnsley10 (approved):
Looks good. Approving but obvs unit tests and coverage need a bit of resolving

PR #1475, opened 2026-06-02
feat/df-1137: Model changes and functionality to use drilldown data in metrics

jbarnsley10:
Pairs with PR https://github.com/DEFRA/forms-audit-api/pull/159

Drilldown data is now stored separate to the overview record, so as not to risk hitting the mongodb doc limit. This PR reads the data from a new endpoint rather than from the original overview report endpoint.

Also fixes a cosmetic bug where the unit 'days' was being incorrectly styled in the 'Average time to publish' tile

### DIFF
--- a/designer/server/src/common/components/metrics-tile/template.njk
+++ b/designer/server/src/common/components/metrics-tile/template.njk
@@ -11,12 +11,12 @@
 <!-- Metrics tile -->
 <div class="{{ tileClasses }}" data-testid="{{ metricTestId }}">
   <div class="app-metric-tile">
-    <div class="gem-c-big-number">
-      <span class="gem-c-big-number__label">{{ params.title }}</span>
+    <div class="app-c-big-number">
+      <span class="app-c-big-number__label">{{ params.title }}</span>
       {% if params.count > 0 and params.drillDown.enabled %}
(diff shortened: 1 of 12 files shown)


designer/server/src/common/components/metrics-tile/template.njk
```diff
@@ -7,18 +7,16 @@
 {% endif %}
 {% set metricTestId = params.testId or ('headline-column-' + (params.title | lower | replace(' ', '-'))) %}
 {% set tileClasses = params.classes if params.classes else "govuk-grid-column-one-half govuk-grid-column-one-quarter-from-tablet govuk-grid-column-one-quarter-from-desktop" %}
+{% set units = '<span class="app-big-number-unit"> ' + params.units + '</span>' if params.units else '' %}
 <!-- Metrics tile -->
 <div class="{{ tileClasses }}" data-testid="{{ metricTestId }}">
   <div class="app-metric-tile">
     <div class="gem-c-big-number">
```
davidjamesstone: I wonder if we should've adopted these styles by replacing the `gem-` prefix with `app-`?

davidjamesstone (approved):

designer/server/src/common/components/metrics-tile/template.njk
jbarnsley10: Changed

PR #1480, opened 2026-06-05
Conditions with Multiple Options

Scullyon:
Allowing multiple options to be chosen for conditions based on the question types:

- Checklist (AND or OR multiple value options together)
- Autocomplete (OR multiple value options together, AND for negative / NOT conditions)
- Radio (OR multiple value options together, AND for negative / NOT conditions)

This new functionality is backwards compatible with the previous 'single option' only functionality.

I have a file that I can share that helps explain the PR - just ping me for it.

Release order:

- Manager
- Plugin
- Runner

will all need to be updated to include the latest Model from this PR before this is released.

image

image

image

image

image

image

image

image


model/src/form/form-definition/index.ts
```diff
@@ -256,10 +260,20 @@ const conditionListItemRefDataSchemaV2 =
         .error(checkErrors(FormDefinitionError.RefConditionListId)),
       itemId: Joi.string()
         .trim()
-        .required()
-        .description('The id of the list item')
-        .custom(listItemIdValidator)
+        .description('The id of the list item (legacy single selection)')
```
davidjamesstone: I wonder if it's better to alter `itemId` rather than introduce a new `itemsId` field?

Types
--------
`itemId: string | string[]`

Schema
-----------
`itemId: joi.array().items(joi.string().uuid()).single()`

davidjamesstone: By using `joi.array.single()` you can safely treat everything as an array once the data has been validated.

designer/server/src/models/forms/editor-v2/condition-helper.js
```diff
@@ -281,13 +282,62 @@ export function buildConditionsFields(
...
+
+  // Single-answer list questions (radio/autocomplete/select) combine multiple
+  // selections with OR for a positive match, but with AND for an "is not" match
+  // (i.e. the answer must not be any of the selected options).
+  const isAndCombination =
+    conditionType === ConditionType.ListItemRef &&
+    !isCheckboxList &&
+    operatorValue === OperatorName.IsNot
+
   // prettier-ignore
```
davidjamesstone, line 303: Can we avoid ignoring prettier?

designer/server/src/models/forms/editor-v2/condition-helper.js
```diff
@@ -281,13 +282,62 @@ export function buildConditionsFields(
...
   const value = 'operator' in item && component.value && operator?.value
       ? buildValueField(conditionType, idx, item, selectedComponent, definition, validation)
       : undefined
 
   // prettier-ignore
```
davidjamesstone, line 308: Ditto

designer/server/src/models/forms/editor-v2/condition-helper.js
davidjamesstone: I've noticed there are a few other `prettier-ignores` in this repo. I'll create a ticket for the others.

Scullyon: Prettier-ignore are existing, right?

model/src/form/form-definition/index.ts
Scullyon: It's a thought, though could it be confusing having multiple items in a singular ID named property. The rest of the code around this wouldn't change as we'd still need to check everywhere whether it was an array or single item (so the new helper would remain).

Ideally, we'd get rid of the `itmeId` property, but even if we migrated all the forms across all environments, would there be a risk where people already have a downloaded JSON form? Then we'd need some additional logic to auto-migrate any uploaded form.

designer/server/src/models/forms/editor-v2/condition-helper.js
davidjamesstone: Doh!

model/src/form/form-definition/index.ts
davidjamesstone: We have some other areas in code where we use a singular name for something that could be 1 or many so I'm not too worried about that. Personally I think it's cleaner in this scenario rather than introducing a new `itemsId` so if it's not a major PITA to refactor then I'd go with that but maybe ask Alex and Jez too see what they think?

davidjamesstone: Plus, we already have this kind of thing with conditions:

```
export type ConditionValueDataV2 =
  | ConditionListItemRefValueDataV2
  | RelativeDateValueDataV2
  | string
  | boolean
  | number
```

where the `value` could be any one of those shapes above, so extending `itemId` to allow it to be 1 or many kind of works (for me!)

Scullyon: I can update it, it's not a problem and I get the usage of union types in other places. It's more of that having a single item is no longer correct, it should always be an array even if it's just a single item. The only reason for have a singular version is for backwards compatibility, and if someone is editing the JSON, it makes it clear what's to expected. However, it's also not ideal having a deprecated property left either, so there aren't any perfect options either way. I'll push an updated version.

davidjamesstone: Resolved with https://github.com/DEFRA/forms-designer/pull/1480/commits/174d52bb97e3a5f4f1af2b640a2b99dcbbc5f47e

davidjamesstone (approved):

PR #1503, opened 2026-06-18
Accessibility - Updates from Main and Maps Updates

Scullyon:
- Fixing Designer to work with ESM maps
- Latest updates from main

### DIFF
--- a/designer/package.json
+++ b/designer/package.json
@@ -39,7 +39,7 @@
     "@defra/forms-engine-plugin": "^4.19.1",
     "@defra/forms-model": "^3.0.681",
     "@defra/hapi-tracing": "^1.28.0",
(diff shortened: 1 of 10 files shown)


designer/package.json
```diff
@@ -39,6 +39,7 @@
     "@defra/forms-engine-plugin": "^4.19.1",
     "@defra/forms-model": "^3.0.681",
     "@defra/hapi-tracing": "^1.28.0",
+    "@defra/interactive-map": "0.0.22-alpha",
```
davidjamesstone: We shouldn't need a dependency on "@defra/interactive-map" - this should come via the plugin

Scullyon: It's required in maps.js. Have added an override to sync the versions.

davidjamesstone (approved):

PR #1508, opened 2026-06-18
feat/df-1164: Add ref num to submission payload (optional)

jbarnsley10:
Allow 'reference number' to be passed in submission payload, so it can be inserted in the CSV file.
Initially 'optional' to avoid any glitches during deployment. To be made mandatory after initial release.

### DIFF
--- a/model/src/form/form-submission/index.ts
+++ b/model/src/form/form-submission/index.ts
@@ -73,6 +73,11 @@ export const formSubmitPayloadSchema = Joi.object<SubmitPayload>()
     sessionId: Joi.string()
       .required()
       .description('User session identifier for tracking and security'),
+    referenceNumber: Joi.string()
+      // TODO - make this required in later releases - optional for now to avoid release glitches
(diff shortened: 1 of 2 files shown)


model/src/form/form-submission/index.ts
```diff
@@ -73,6 +73,11 @@ export const formSubmitPayloadSchema = Joi.object<SubmitPayload>()
     sessionId: Joi.string()
       .required()
       .description('User session identifier for tracking and security'),
+    referenceNumber: Joi.string()
+      // TODO - make this required in later releases - optional for now to avoid release glitches
```
Scullyon, line 77: Is there a ticket for this?

Scullyon (approved):

model/src/form/form-submission/index.ts
jbarnsley10: [DF-1177](https://eaflood.atlassian.net/browse/DF-1177)

PR #1509, opened 2026-06-18
Remove prettier-ignore statements

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-1169

### DIFF
--- a/designer/server/src/models/forms/editor-v2/guidance.js
+++ b/designer/server/src/models/forms/editor-v2/guidance.js
@@ -182,8 +182,13 @@ export function guidanceViewModel(
   const sectionInfo = page
     ? getSectionForPage(definition, page, metadata.slug)
     : undefined
-  // prettier-ignore
-  const previewModel = getGuidancePreviewModel(page, previewPageUrl, guidanceTextVal, sectionInfo)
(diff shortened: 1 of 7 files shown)


designer/server/src/models/forms/editor-v2/guidance.js
```diff
@@ -182,8 +182,13 @@ export function guidanceViewModel(
   const sectionInfo = page
     ? getSectionForPage(definition, page, metadata.slug)
     : undefined
-  // prettier-ignore
-  const previewModel = getGuidancePreviewModel(page, previewPageUrl, guidanceTextVal, sectionInfo)
+  const previewModel = getGuidancePreviewModel(
```
Scullyon, line 185: I actually prefer the single line version.

Scullyon (approved):
In a lot of places, I prefer the single line version and think some of the changes to appease prettier make the code slightly harder to read. Perhaps worth looking at the prettier config at some point?

designer/server/src/models/forms/editor-v2/guidance.js
davidjamesstone: Same here - yeah let's review the prettier config - I have no strong opinions but I'd just want to avoid `prettier-ignore`

PR #1523, opened 2026-07-16
Add map layers advanced configuration for map-based components

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-1218

image

### DIFF
--- a/designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
+++ b/designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
@@ -138,6 +138,13 @@ export function getAdditionalOptions(payload) {
         payload.countries[0] !== 'any'
     },
     {
+      key: 'mapLayers',
+      getValue: () => ({ sssi: payload.mapLayers?.includes('sssi') ?? false }),
+      shouldInclude: () => {
+        return Array.isArray(payload.mapLayers) && payload.mapLayers.length
+      }
+    },
+    {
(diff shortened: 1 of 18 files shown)


designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
```diff
@@ -137,6 +137,13 @@ export function getAdditionalOptions(payload) {
         payload.countries.length === 1 &&
         payload.countries[0] !== 'any'
     },
+    {
+      key: 'mapLayers',
+      getValue: () => ({ sssi: payload.mapLayers?.includes('sssi') ?? false }),
```
jbarnsley10, line 142: Could use function `isCheckboxSelected()` as in line 96

jbarnsley10 (approved):
Looks good. One minor suggestion. Approving in principle.

designer/server/src/models/forms/editor-v2/advanced-settings-helpers.js
davidjamesstone: Not sure this will work here actually. I've made `mapLayer` extensible, the thinking being that `sssi` might just be the first layer we support. The `payload.mapLayers` is an array here, not "true" or "Y".

jbarnsley10: Sure, I hadn't appreciated that

PR #1526, opened 2026-07-24
feature/DF-1220: Form submission page

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-1220

image

### DIFF
--- a/designer/server/src/models/submission/index.js
+++ b/designer/server/src/models/submission/index.js
@@ -6,7 +6,8 @@ import {
   hasRepeater,
   replaceCustomControllers
 } from '@defra/forms-model'
-import { format } from 'date-fns'
+
+import { format } from '~/src/models/forms/history-date-utils.js'
 
 /**
  * @param {FormAdapterSubmissionMessagePayload} submission - the form submission record
--- a/designer/server/src/models/submission/index.js
+++ b/designer/server/src/models/submission/index.js
@@ -19,7 +19,7 @@ import { format } from '~/src/models/forms/history-date-utils.js'
 function processSectionComponent(component, page, context, repeat) {
   let value = ''
   let actions
-  const { formModel, data, translator, referenceNumber } = context
+  const { formModel, submission, translator, referenceNumber } = context
   const field = formModel.componentMap.get(component.name)
 
   if (!(field instanceof FormComponent)) {
@@ -27,11 +27,14 @@ function processSectionComponent(component, page, context, repeat) {
   }
 
   if (component.type === ComponentType.FileUploadField) {
-    if (component.name in data.files) {
-      const files = /** @type {any} */ (data.files[component.name])
+    if (component.name in submission.data.files) {
+      const files = submission.data.files[component.name]
 
       if (Array.isArray(files) && files.length) {
-        value = field.getDisplayStringFromFormValue(files, translator)
+        value = field.getDisplayStringFromFormValue(
+          /** @type {any} */ (files),
+          translator
+        )
         actions = [
           {
             href: `/files-download/${referenceNumber}`,
@@ -48,7 +51,7 @@ function processSectionComponent(component, page, context, repeat) {
       }
     }
   } else {
-    const source = repeat?.item ?? data.main
+    const source = repeat?.item ?? submission.data.main
 
     if (component.name in source) {
       value = field.getDisplayStringFromFormValue(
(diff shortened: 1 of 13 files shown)


designer/server/src/views/submission/index.njk
```diff
@@ -0,0 +1,51 @@
...
+
+{% block content %}
+  {% call appPageBody({
+    heading: {
+      text: pageHeading,
+      size: "large",
+      caption: caption
+    }
+  }) %}
+    <p class="govuk-body">{{ lede }}</p>
```
jbarnsley10: Is this line meant to be here?

designer/server/src/views/submission/index.njk
```diff
@@ -0,0 +1,51 @@
...
+    {% for summary in summaries %}
+      {% if summary.section %}
+        <h2 class="govuk-heading-m">{{ summary.section }}</h2>
+      {% endif %}
+
+      {# Preprocess #}
+        {% set rows = [] %}
```
jbarnsley10, line 29: Is it better to have these lines of code in the model so that can be unit tested? Or are you able to create a fully-=featured snapshot to prove it's all working?

designer/server/src/views/submission/index.njk
davidjamesstone: Yes - this writes out the "Submitted at dd MMMM YYYY" line although it's badly named - I'll change it from `lede`

designer/server/src/models/submission/index.js
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * @param {FormAdapterSubmissionMessagePayload} submission - the form submission record
+ * @param {FormDefinition} definition - the form definition
+ */
+export function submissionViewModel(submission, definition) {
+  const { meta } = submission
+  const { referenceNumber } = meta
+  const fixedDefinition = replaceCustomControllers(definition)
+  const formModel = new FormModel(fixedDefinition, { basePath: '' })
+  const translator = formModel.createTranslator()
```
jbarnsley10, line 173: Should we be resolving what language to use?
If translation is not needed, do we even need to create a translator (c.f. line 186)

designer/server/src/models/submission/index.js
```diff
@@ -0,0 +1,220 @@
...
+    translator,
+    submission,
+    formModel,
+    referenceNumber
+  }
+
+  ;[...sections, undefined].forEach((section) =>
+    processSection(section, context)
+  )
+
```
jbarnsley10, line 195: I had to lookup what 'lede' referred to (apparently slang for 'lead'). Can we use a var name that might make a bit more sesne to the uninitiated (like me - haha)

designer/server/src/views/submission/index.njk
davidjamesstone: I do have a fully featured snapshot that asserts the screenshot above, including all component types.

The reason it has to be in njk is because we're including a `govukSummaryList` inside another `govukSummaryList` value (for repeaters).

I wanted to avoid generating the HTML inside JS so had to build up these "child summaries" in the template file.

jbarnsley10 (commented):
I'm not sure if we need any translation capability here or not, since it's in designer. So I may have misunderstood the requirements.

designer/server/src/models/submission/index.js
```diff
@@ -0,0 +1,220 @@
...
+  const lede = `Submitted on ${format(meta.timestamp, "dd MMMM yyyy 'at' h:mm")}`
```
davidjamesstone: Maybe eventually but like you say, it's in designer so not a requirement yet but it might be we want a toggle to view the submission in the language it was submitted in

designer/server/src/models/submission/index.js
davidjamesstone: The paragraph class `govuk-body-l` used to be called `govuk-body-lede`.  `lede` is an old term but still used in typesetting/newspaper jargon. I'll rename it as it's not very useful here

jbarnsley10 (approved):
Looks great

designer/server/src/models/submission/index.js
davidjamesstone: We now need a translator when calling `field.getDisplayStringFromFormValue(value, translator)` so I'm just creating and passing in the default (English) one.

